PR #523, opened 2014-12-16
Separate plugin config file

neilstuartcraig:
Hi Evan

Here's my PR WRT https://github.com/evantahler/actionhero/issues/512. It's a branch obviously, I thought that'd likely be best but I can merge into `master` if you prefer.

I hope i've done the right thing but please let me know if you don't think I have. It's tricky to test without being able to install via NPM but from what I can see, the plugins are loading correctly.

I wasn't sure whether to add to `package.json` `install` string or not but this PR includes the option for an argument on the `generate` method to add plugins to the new `config/plugins.js` file. The arg is `--plugins` and just uses the existing `binary.argv` object. My thinking is that this doesn't break any existing script people may be using and does not add plugins by default but does allow programmatic addition of plugins via a little shell scripting or similar.

Hope that's all OK anyway. If so, I can perhaps update the `gh-pages` branch to include the info on the flag and the config file.

Cheers
Neil

### DIFF
--- a/bin/methods/generate.js
+++ b/bin/methods/generate.js
@@ -41,22 +41,23 @@ exports.generate = function(binary, next){
   documents.readmeMd    = String(fs.readFileSync(binary.paths.actionheroRoot + '/bin/templates/README.md'));
 
   // Add matching package.json dependencies to the dedicated plugins config file
+  var dep;
   var pluginsArrayContents="";
-  if(typeof(JSON.parse(documents.packageJson).dependencies)==="object")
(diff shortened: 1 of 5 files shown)


evantahler:
Thanks for your work!

No, I don't have a PR guide... feel free to make one!  In general, I follow the rule that "contain everything you need for the future... and minimize the code to do so".  So in your example, I think this one PR has the basics with the generate flag.  The other bullets you mention can be updates to the feature later on. 

Regarding tests for the generators, I wrote this story to capture the fact that we have no tests :( 

https://github.com/evantahler/actionhero/pull/523#issuecomment-68306025

PR #532, opened 2014-12-31
add generateMessagePayload in chatroom

vahid-sohrabloo:
Add generateMessagePayload to customize incoming message


evantahler:
You could add it here.  There are already clients set up and connected to a room https://github.com/evantahler/actionhero/blob/master/test/core/actionCluster.js#L119-L129

evantahler:
You'll need to update your branch on master... the chat system has seen some refactoring as of late

vahid-sohrabloo:
Sorry for my delay.
This is my first Test in nodejs :smile:

evantahler:
Thanks for working on the tests!
I ended up copying your code into a new branch https://github.com/evantahler/actionhero/pull/546 rather than make you go back and make some small changes.  Generally, you want to to test this kind of thing agianst a generic testServer, as this method applies to all transports which can chat... not just websockets.

Thanks for your work!

PR #535, opened 2015-01-02
Ability to configure initializer load precedence (WIP)

evantahler:
1. we need to move the redis connection in that initializer to the `_start()` method so it can be modified like all other initializers (new-relic requires this for example)
2. we should removed the "orderedInitializers" section from the boot process so developers can inject their own initializers before if needed
3. this requires some sort of priority system for initializers. 

- we can append initializers via a method `api.addInitializer(object, priority)`
- we could add another property to the initializer, like `_priority` which will be used to determine load order

---

We've gone with the method which changes initializers in the following way:

```javascript
module.exports = {
  loadPriority:  1000,
  startPriority: 1000,
  stopPriority:  1000,
  initialize: function(api, next){
    api.%%name%% = {};

    next();
  },
  start: function(api, next){
    next();
  },
  stop: function(api, next){
    next();
  }
}
```

Note that initializers now:
  - The module is a single export (`module.exports`)
  - There are 3 lifecycle events that initializers can respond to: `initialize`, `start`, and `stop`.
    - The `_start` and `_stop` methods have been moved to the top level (and renamed to `start` and `stop`), along with the `initialize` method.
    - Each of these methods is passed the `api` object and a callback. 
  -  (optionally) have a `loadPriority` (system levels are > 1000)
  -  (optionally) have a `startPriority` (system levels are > 1000)
  -  (optionally) have a `stopPriority` (system levels are > 1000)


actionhero.js
```diff
@@ -43,95 +44,142 @@ actionhero.prototype.initialize = function(params, callback){
   self.startingParams = params;
   self.api._startingParams = self.startingParams;
 
-  // run the initializers
-  var orderedInitializers = {};
+  self.api.initializerDefauls = {
```
jacobfike: Missing the "t" in Defaults here.

evantahler: Thanks!

PR #541, opened 2015-01-06
Allow the use of multiple config specifications.

crrobinson14:
This PR allows the use of multiple config paths. A typical use-case would be to provide default config settings in a project, with "local" settings that apply only to a specific server. This makes it easy to keep database passwords, API keys, and other security-sensitive information out of your repository, or to adjust server-specific settings such as which nodes will run background tasks vs. Web services.

This change preserves the existing load order for ActionHero configs:

1. The project's 'config' folder, if exists
1. `actionhero --config=PATH` parameters. More than one --config value may now be provided.
1. `ACTIONHERO_CONFIG=PATH npm start` style environment variables.

In addition, config paths may include more than one path, separated by commas. This is handy when using the environment-variable style, e.g. `ACTIONHERO_CONFIG=config,local-config npm start`.

As with standard ActionHero configs, override files do not need to re-specify every key from the default file - just the keys being overridden.

### DIFF
--- a/initializers/configLoader.js
+++ b/initializers/configLoader.js
@@ -46,19 +46,16 @@ module.exports = {
       api.env = process.env.NODE_ENV;
     }
 
-    // We now support multiple configuration paths. Repeated uses of --config will add paths to examine. But note that
-    // the old behavior <10.x was that specifying --config would OVERRIDE the default path (the configs that ship with
-    // AH). To preserve that behavior, we use the AH default here as before, but overwrite below if config is specified.
+    // We support multiple configuration paths as follows:
     //
-    // Load order:
-    // 1. Project 'config' folder, if exists
-    // 2. "actionhero --config=PATH"
-    // 3. "ACTIONHERO_CONFIG=PATH npm start"
+    // 1. Use the project 'config' folder, if it exists.
+    // 2. "actionhero --config=PATH1 --config=PATH2 --config=PATH3,PATH4"
+    // 3. "ACTIONHERO_CONFIG=PATH1,PATH2 npm start"
     //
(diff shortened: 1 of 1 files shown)


crrobinson14:
Seems like the tests that fail are unrelated. I'm happy to make any changes required if necessary, though.

jtmarmon:
Perhaps I'm not grepping the purpose of your PR, but it seems a lot cleaner and simpler to use env variables for this stuff. It makes more sense to me to have config files which container their own logic and just let you use env variables for the secure configuration parts. That's what I've been doing at least.

crrobinson14:
That's not the best strategy for the kinds of material this PR lets you provide overrides for. Consider:
1. Environment variables are exposed in a number of ways - they're not great places to put passwords.
2. Some types of config elements require access to the "api" object to make decisions (which is why it's passed to each callback in the first place). Environment setup is too early to use PART of the startup to make the next decision.
3. Environment variables are a tedious place to put settings if you need many of them. It's fine for a port or two... but not for an entire MySQL/Redis config setup with host, port, user, pass, pooling, etc.

What this PR does is allow you to use the standard 'config' folder, as you do today, but in addition you can provide something like a 'local-config' folder that would provide overrides. This makes it very easy to put that folder into your .gitignore and thus avoid accidentally (or deliberately) committing passwords, API keys, and other things to your repository. This is a common best practice and is very important for enterprise software projects. You can look at Drupal, Loopback.io, Django, and well, most other major frameworks for examples of this pattern.

It's also optional. The old functionality still works as-is - this just gives you an extra level of control when you want it.

jtmarmon:
fair nuff. consider me educated :smile:

evantahler:
Looks great!

evantahler:
Added to the v10 release & release notes https://gist.github.com/evantahler/f38ee2834d79f52f6996

PR #542, opened 2015-01-07
Fix typo in initializers docs

jtmarmon:


jtmarmon:
just noticed that this typo contradicted the point of that section so i figured i'd fix it

evantahler:
Ha! Thanks.  This will all be changing with the new release of course.

PR #548, opened 2015-01-11
mongo initializer example

sallespro:
much to thank, not to say


evantahler:
Thanks! looks like I need to update the spelling in the navigation too :D

PR #559, opened 2015-01-13
tests for the actionhero binary

evantahler:


evantahler:
Would close https://github.com/evantahler/actionhero/issues/539

PR #562, opened 2015-01-15
Make HTTP action callbacks use the same param signature as WebSocket calls

crrobinson14:
...lbacks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,9 @@ services:
 before_script:
   - npm install -g grunt-cli
 
+after_script:
+  - npm run-script coverage
+
 notifications:
   webhooks:
     urls:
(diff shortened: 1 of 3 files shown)


crrobinson14:
Travis did the same thing to me last time. The tests all pass, but one node runs too slowly and times out. Just luck? Committing again to bump it through.

client/actionheroClient.js
```diff
@@ -155,9 +155,9 @@ ActionheroClient.prototype.actionWeb = function(params, callback){
     if(xmlhttp.readyState === 4){
       if(xmlhttp.status === 200){
         var response = JSON.parse(xmlhttp.responseText);
-        callback(null, response);
-      }else{
-        callback(xmlhttp.statusText, xmlhttp.responseText);
+        callback(response);
+      } else {
+        callback(xmlhttp.responseText);
```
evantahler: There's probably something to do here.  We generally know that in this case `xmlhttp.responseText` will be an error, so to present a homogenous response object, you probably want to return `callback({error: xmlhttp.responseText})`

crrobinson14:
Closing this PR because it's a little awkward in Github to have overlapping PRs from a single fork. Let's discuss the other one, about more flexible middleware declarations, first, then revisit this.

PR #568, opened 2015-01-21
SVG image update

jbender:
Quick update to the readme to provide more constant formatting and improved rendering for higher-dpi screens


evantahler:
Much better looking.  Thanks!

PR #570, opened 2015-01-24
Add support for per-action middleware

crrobinson14:
This PR adds support for more generic middleware functionality. I didn't write a full gh-pages document update yet (I wanted feedback on the PR itself yet) - but I will if this is accepted. In the meantime, this allows middleware authors to call a new function to register their modules:

```javascript
api.actions.addMiddleware('jwtSession', {
  preprocess: function(connection, actionTemplate, next) {
    var config = actionTemplate.middleware.preprocess.jwtSession || { default: 1 };
    next(connection, true);
  }
});
```

Middleware modules may register for preprocess, postprocess, or both. Future enhancements could allow many more hooks, obviously, so I kept it generic, as a collection of hooks to provide.

Nothing further happens yet. Middleware that want to also provide global functionality for all requests can continue to do so using the usual `addPreProcessor()/addPostProcessor()` calls.

To use these enhanced middleware modules, actions add a new configuration key to their metadata blocks:

```javascript
middleware: {
    preprocess: {
        jwtSession: {
            loadFullUser: true
        },
        quotaManager: {}
    },
    postprocess: {
        requestAuditor: {}
    }
},
```

This allows actions to be declarative about the middleware that should be run before/after they're called. This system is opt-in since middleware can still provide global hooks for all actions. But using this new mechanism, you can define in the action whether something like a user session loader needs to run (yes for authenticated requests, no for anonymous requests). session, quota, and audit handlers were the three use-cases I imagined (actually, that I HAD) but I'm sure there are many more.

I've deliberately written the code to more or less parallel what's currently done for the global handlers, so it's a bigger diff than it needs to be. There's some duplication of code in how the array of async callbacks is constructed - if this PR is accepted I can refactor them to make the code more elegant.


crrobinson14:
Oh, note that I illustrate how per-action configs can be handled, but didn't actually provide a sample. `loadFullUser` in the example above would be `config.loadFullUser` in the handler shown.

neilstuartcraig:
Hi guys. 

I'm away in the mountains at the moment so don't gave full access to check this but I have a plugin which does something similar. Maybe the two ideas could merge?

My basic plan was to allow glob-able action name patterns and versions mapped to middleware. It's in this repo at the moment: https://github.com/neilstuartcraig/TDPAHDynamicPrePostProcessorLoader

I'd be very interested to discuss if you guys think this general idea is a go-er. 

Cheers
Neil

jtmarmon:
awesome. being able to write middleware for specific actions is definitely something I've needed (e.g. rate limiting certain actions per-ip)

crrobinson14:
@neilstuartcraig Your module is very interesting - thanks for contributing it! I'm not sure if they can be MERGED though. I think each contribution adds something different. This PR focuses on allowing actions to be declarative about the middleware they require via settings in their meta-data blocks. Yours focuses on allowing _middleware_ to be much more flexible. They're pretty different goals because one puts the burden on the action developer and the other puts it on the middleware developer.

In fact, it seems to be that some people might want BOTH options in their projects. Some kinds of middleware are more utility-oriented, generically useful across a mix of actions. Your plugin is probably the best choice for those because if you want to apply 12 different middlewares across a mix-and-match of actions, it would be kind of annoying to have to specify them in EVERY action file. Your own examples are perfect use-cases - capitalization, translation, and filtering tools.

Mine is more focused on cases where you have fewer choices made on broader bases, and especially where configs might change from action to action. Session managers are a good example, and so are auditing and rate-limiting mechanisms. For those, it would be super confusing to have a huge bucket of config blocks somewhere outside the actions forcing you to look at two files to figure out what an action is going to do...

I would guess that they could easily be made compatible, though. All we'd need is a call to api.actions.addMiddleware here, where addPreprocessor is also done:
https://github.com/neilstuartcraig/TDPAHDynamicPrePostProcessorLoader/blob/master/TDPAHDynamicPrePostProcessorLoader.js#L35

If that was done, developers could use both options simultaneously!

neilstuartcraig:
Yeah, it's something I wrote quite some time ago and then updated to ah v9 more recently. 

I think mine could quite easily exist as a plugin and that might be the simplest way to use it. I need to tidy it quite a bit but would really welcome any thoughts and suggestions. 

That said, I have no objection in some or all of my plugin becoming part of the ah core but that depends very much on it being generic enough and appropriate of course. 

Cheers

actions/middleware.js
```diff
@@ -0,0 +1,17 @@
+exports.middleware = {
```
evantahler: Please don't check in an action that won't actually run.  The middleware referenced here aren't present in the project.

evantahler:
In general, I like this a lot.  I think this better in a few ways:

- having 1 generic "add middleware" that can take an object which may have both a `pre` and `post` block is better than 2 methods to add only a `pre` or `post` block.
- I think for the majority of folks, choosing the middleware on the action will be more useful than creating global middleware.

That said, I can still think of global (or used most-of-the-time middleware), like a rate-limiter or auth plugins, would be usefule.  Perhaps `api.actions.addMiddleware`'s arguments are extended to look like:

```javascript
api.actions.addMiddleware('jwtSession', {
  preprocess: function(connection, actionTemplate, next) { 
     // ... 
     next(connection, true);
  },
  postprocess: function(connection, actionTemplate, toRender, next) { 
     // ... 
     next(connection, toRender);
  },
  global: true // <-- here
});
```  

Where `global = true` can be set to toggle this middleware on for all actions.  Then, in the action's defintion regarding middleware you can do:

```javascript
middleware: {
    preprocess: {
        jwtSession: {
            loadFullUser: true
        },
        quotaManager: {}
    },
    postprocess: {
        requestAuditor: {}
    },
    ignoreGlobals: true // <--- here
},
```

I am 100% OK with this PR removing the old middleware stack in favor of the new style if the global considerations are accounted for.  I do not want to have a state in the project with both "styles" of middleware are supported (because it is confusing and 2x the work to manage)

actions/middleware.js
crrobinson14: It's used by the test case. Is there another way to do the test case (since the other actions don't need middleware)?

crrobinson14:
That makes perfect sense to me. I can adjust it tonight. I assume that means you'd want to put this in 11x since it's a breaking change... Or do you want to keep the add[Pre|Post]Processor and just have them be basically a polyfill from the old behavior to the new?

evantahler:
Nice! Feel free to break anything you want :D 
I've started the [v11 milestone](https://github.com/evantahler/actionhero/milestones/v11.0.0), and this is part of it!

evantahler:
For the test suite, you can always add actions within: https://github.com/evantahler/actionhero/blob/master/test/servers/web.js#L148-L172

crrobinson14:
OK, I switched the tests to the other method, but just to confirm, you'd want this to remove add[Pre|Post]Processor? The only thing that uses them in the framework is the test suite. You have a lot more tests for the generic functionality than I put into my case because I hit a lot of combinations in one pass.

For example, the current test list has a check to be sure it can run a pre-processor. Then it tests that it can run pre-processors in the right order. I didn't do the first test against the middleware aspect because by proving you can run them in the right order, you prove that you can run them at all. :) But that's a personal preference - I know some people prefer more, smaller tests to help identify where failures happen when they do. Let me know what you're comfortable with here.

crrobinson14:
FYI this is just me rebasing these changes on top of PR #571 and getting rid of the dummy action per Evan's request above. It does not (yet) remove add[Pre|Post]Processor.

evantahler:
@crrobinson14 What do you want to do with this PR?  I think the `global` flag should be added here... and we should remove the old pre/post processors all in one go.

crrobinson14:
@evantahler Makes sense to me, but we're pre-launch on our project and I'm super tight on time for a bit. It might be next week before I can complete this.

evantahler:
no worries!  Just checking in.

crrobinson14:
@evantahler I'm ready to get back into this and just wanted to review where we stand so we can get it finalized.

Just to recap, you're cool with making middleware more expressive and declarative, but you want to go all-in and remove the old addPreprocessor/addPostprocessor options as well? I just want to confirm this because it looks like there's increasing interest from some other contributors for making plugins, and I didn't want this to be a blocker for them.

evantahler:
Yeah, feel free to break everything (in order to make it better)!

v12 will be a breaking change already because of this already: https://github.com/evantahler/actionhero/pull/615.  

Please build this PR against the v11 branch: https://github.com/evantahler/actionhero/pull/621.  This means that the API changes to:

```javascript
api.actions.addMiddleware('jwtSession', {
  preprocess: function(data, next) { 
     // ... 
     next(error);
  },
  postprocess: function(data, next) { 
     // ... 
     next(error);
  },
  global: true // <-- here
});
```

where 

```javascript
data = {
  connection: connection,
  action: 'randomNumber',
  toProcess: true,
  toRender: true,
  messageCount: 123,
  actionTemplate: {...}
  params: { action: 'randomNumber', apiVersion: 1 },
  actionStartTime: 123,
  response: {},
}
```

So:
-  `toRender` is now `data.toRender`
-  `actionTemplate` is now `data.actionTemplate`

evantahler:
I got bored waiting... so I did this myself: https://github.com/evantahler/actionhero/pull/625

evantahler:
Everyone move the conversation over here! https://github.com/evantahler/actionhero/pull/625

evantahler:
... and the ideas in this thread are awesome.  Please check https://github.com/evantahler/actionhero/pull/625 to make sure that I actually captured everyone's thoughts!

PR #571, opened 2015-01-27
reconnect try

evantahler:


evantahler:
This PR fixes a race condition which would result in the wrong messageIds being sent from the server (compared to what the client expected).  This only occurred when you joined a room before primus was "fully" connected, or you rapidly connected and disconnected. 

Reported by @crrobinson14

PR #573, opened 2015-01-30
fix route regexp

keifukuda:
Regular expression match of routing, was not able to use the grouping.

### DIFF
--- a/initializers/routes.js
+++ b/initializers/routes.js
@@ -38,10 +38,10 @@ module.exports = {
       var matchParts = match.split('/');
       var regexp = '';
       var variable = '';
-      
+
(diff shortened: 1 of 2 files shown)


evantahler:
Can you provide an example of what this fixes?

keifukuda:
Including the parentheses is a regular expression.

```javascript
path: '/login/:userID(^foo(bar|baz)$)'
```

evantahler:
Awesome!

Please add a test for this (probably in this section) https://github.com/evantahler/actionhero/blob/master/test/servers/web.js#L612-L627

evantahler:
Thanks!

PR #581, opened 2015-02-11
resolve issue with additional "." in initalizer path

S3bb1:
if your current folder contains any "." in it ... all initalizers will break. this PR uses a nodejs path function to solve this.

### DIFF
--- a/actionhero.js
+++ b/actionhero.js
@@ -93,7 +93,7 @@ actionhero.prototype.initialize = function(params, callback){
       )
     ).forEach(function(f) {
       var file = path.normalize(f);
-      var initializer = f.split('.')[0];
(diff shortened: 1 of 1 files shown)


evantahler:
nice.

jacobbubu:
Could make it more consistent via `path.extname`:

```js
var initializer = path.basename(f).split('.')[0];
if (path.extname(file) === '.js') {
     ...
}
```

S3bb1:
yeah, looks good maybe we could replace the "f" variable with the "file" variable in the basename parameter too.

``` javascript
var initializer = path.basename(file).split('.')[0];
if (path.extname(file) === '.js') {
     ...
}
```

evantahler:
You can update this branch and send a new PR from it!

PR #583, opened 2015-02-11
Fix unexpected param filtering behavior - resolves #582

### LINKED ISSUE #582: Unexpected parameter scrubbing behavior
Parameters seem to be filtered based on whether or not they've been defined in *any* action rather than whether they've been defined in the target action or are global safe parameters.

dangerdespain:
Modify the reduceParams function of the actionProcessor initializer to allow params from the globalSafeParams array rather than the postVariables array

### DIFF
--- a/initializers/actionProcessor.js
+++ b/initializers/actionProcessor.js
@@ -176,7 +176,7 @@ module.exports = {
       if(api.config.general.disableParamScrubbing !== true){
         for(var p in self.connection.params){
           if(
-              api.params.postVariables.indexOf(p) < 0 &&
+              api.params.globalSafeParams.indexOf(p) < 0 &&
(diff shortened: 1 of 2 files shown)


evantahler:
This is a great test.  Can you add a test or 2 showing that only params allowed per action are passed into the action?  

Probably belongs in `test/core/api.js`

dangerdespain:
I've added a test to check that parameter filtering. More specifically, I used a parameter defined in a different action (sleepDuration from actions/sleepTest.js) to ensure that parameters aren't leaking in from other actions' parameter definitions as was the case pre-fix.

evantahler:
awesome.  Thanks!

dangerdespain:
Sure thing!

PR #588, opened 2015-02-15
Add invalidParams message handler

rootStar-lock:
We need to handle invalid params and return custom structures, not only strings.


evantahler:
Awesome!

rootStar-lock:
:thumbsup: and sorry for EOLs)

PR #594, opened 2015-02-23
Formal EventEmitter inheritance for AH client

evantahler:
iojs v1.3.0 exposed a bug where (in testing) the AH websocket client lib could have errors with multiple clients.  We were not 'formally' declaring the client as extension of the EventEmmiter, rather just hijacking the prototype.


evantahler:
This will fix the broken io.js tests on master

Issue #587, opened 2015-02-13
Overly verbose debug messages from resque


gcoonrod:
Is this really necessary? What value does this provide at a debug level? IMO, these kinds of messages should be at a trace level.

```
2015-02-13 09:46:31 - debug: [10.137.169.221actionhero-10.137.169.221] worker: polling default {"workerId":1}
2015-02-13 09:46:31 - debug: [10.137.169.221actionhero-10.137.169.221] worker: paused {"workerId":1}
2015-02-13 09:46:31 - debug: [10.137.169.221actionhero-10.137.169.221] resque scheduler polling
2015-02-13 09:46:32 - debug: [10.137.169.221actionhero-10.137.169.221] worker: polling default {"workerId":1}
2015-02-13 09:46:32 - debug: [10.137.169.221actionhero-10.137.169.221] worker: paused {"workerId":1}
2015-02-13 09:46:32 - debug: [10.137.169.221actionhero-10.137.169.221] resque scheduler polling
2015-02-13 09:46:33 - debug: [10.137.169.221actionhero-10.137.169.221] worker: polling default {"workerId":1}
2015-02-13 09:46:33 - debug: [10.137.169.221actionhero-10.137.169.221] worker: paused {"workerId":1}
2015-02-13 09:46:33 - debug: [10.137.169.221actionhero-10.137.169.221] resque scheduler polling
2015-02-13 09:46:34 - debug: [10.137.169.221actionhero-10.137.169.221] worker: polling default {"workerId":1}
2015-02-13 09:46:34 - debug: [10.137.169.221actionhero-10.137.169.221] worker: paused {"workerId":1}
2015-02-13 09:46:34 - debug: [10.137.169.221actionhero-10.137.169.221] resque scheduler polling
2015-02-13 09:46:35 - debug: [10.137.169.221actionhero-10.137.169.221] worker: polling default {"workerId":1}
2015-02-13 09:46:35 - debug: [10.137.169.221actionhero-10.137.169.221] worker: paused {"workerId":1}
2015-02-13 09:46:35 - debug: [10.137.169.221actionhero-10.137.169.221] resque scheduler polling
2015-02-13 09:46:36 - debug: [10.137.169.221actionhero-10.137.169.221] worker: polling default {"workerId":1}
2015-02-13 09:46:36 - debug: [10.137.169.221actionhero-10.137.169.221] worker: paused {"workerId":1}
[...]

PR #595, opened 2015-02-24
Fixes resque verbose output

kytwb:
See issue #587.


evantahler:
Thanks for doing this! There's probably a few other things that should be this new `trace` level:
- Loading files: https://github.com/evantahler/actionhero/blob/master/actionhero.js#L105
- Logging about the Logger: https://github.com/evantahler/actionhero/blob/master/initializers/logger.js#L48
- Route Info: https://github.com/evantahler/actionhero/blob/master/initializers/routes.js#L105 and https://github.com/evantahler/actionhero/blob/master/initializers/routes.js#L117

kytwb:
Done. Let me know if you see more or feel free to merge.

evantahler:
Looks like you haven't actually tried to log something at the `trace` level, it blows us: We need to tell winston the CLI color for our new level, so the logger config block becomes: 

```js
api.logger = new (winston.Logger)({
      levels: {
        emerg: 8,
        alert: 7,
        crit: 6,
        error: 5,
        warning: 4,
        notice: 3,
        info: 2,
        debug: 1,
        trace: 0
      },

      colors: {
        trace: 'magenta',
        input: 'grey',
        verbose: 'cyan',
        prompt: 'grey',
        debug: 'blue',
        info: 'green',
        data: 'grey',
        help: 'cyan',
        warn: 'yellow',
        error: 'red'
      },

      transports: transports
    });
```

we had been previously lucking out because the names of levels were ones Winston had preset as defaults, but `trace` is new.

evantahler:
This is going to be reverted via https://github.com/evantahler/actionhero/pull/712

PR #599, opened 2015-02-24
Add more docs around action inputs

submitteddenied:
Add some more docs for action inputs, explicitly showing that they are optional.


evantahler:
great!

PR #601, opened 2015-02-24
Pass through task documentation

submitteddenied:
Added documentation for a tasks `next` callback.
Fix `api.tasks.allWorkingOn` docs.
Make examples consistent by removing parameters to `next` callback.

### DIFF
--- a/docs/core/tasks.md
+++ b/docs/core/tasks.md
@@ -56,13 +56,14 @@ To work these tasks, you need to run actionhero with at least one `taskProcessor
 
 If you are enqueuing delayed or periodic tasks, you also need to enable the scheduler.  This is a part of actionhero that will periodically check the delayed queues for jobs that are ready to work now, and move them to the normal queues when the time comes.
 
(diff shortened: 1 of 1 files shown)


submitteddenied:
I'd like to sweep back through here again and add more documentation/examples for the various queue inspection methods, but I think that's a PR for another day :smiley:

docs/core/tasks.md
```diff
@@ -138,7 +147,7 @@ exports.sayHello = {
   frequency:     1000,
   run: function(api, params, next){
     api.log("hello")
-    next(null, true);
+    next();
```
evantahler: There is a response signature here, and it is (error, data).  Data doesn't do much (just gets logged), but if you return an `error`, the job will be marked as failed

evantahler:
Thanks for all your work!

Regrading the queue methods, you can link folks here: https://github.com/taskrabbit/node-resque#queue-managment

docs/core/tasks.md
submitteddenied: I do mention that higher up on Line 111, do you think an example of a job that fails (like a divide by with a param 0?) might be a better demonstration of failing a job like this?

evantahler: Oh, I'm just noting that in documentation it's probably valuable to show all possible arguments, even if they are optional.

submitteddenied:
How about this case, where the `sendWelcomeEmail` task passes the error through to fail the job if `sendEmail` fails. This also puts the example of failing a task close to where it is documented.

evantahler:
Oh yeah! That should be updated to pass the `err` to next() as well.  Good call/catch

PR #604, opened 2015-02-26
Refactor route creation and expose api.routes.registerRoute

submitteddenied:
This moves the actual creation of routes to a single place and exposes a method on `api.routes` called `registerRoute(method, path, action)`.

I'm planning to use the `registerRoute` method to add routes to an application from an initializer (in a plugin), hopefully it's useful elsewhere too :)

### DIFF
--- a/initializers/routes.js
+++ b/initializers/routes.js
@@ -73,9 +73,15 @@ module.exports = {
       return response;
     }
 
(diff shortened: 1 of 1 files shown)


evantahler:
Fine by me!

PR #610, opened 2015-03-05
This actually needs to be `undefined`

submitteddenied:
Just a simple correction of documentation here.

It would also be totally reasonable to change this back to `null` and correct where the config is used in `servers/websocket.js:166`.

### DIFF
--- a/config/servers/websocket.js
+++ b/config/servers/websocket.js
@@ -1,23 +1,23 @@
 // Note that to use the websocket server, you also need the web server enabled
 
-exports.default = { 
+exports.default = {
   servers: {
     websocket: function(api){
       return {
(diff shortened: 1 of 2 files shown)


submitteddenied:
Ah, I see my editor has taken liberties with your whitespace too. Sorry about the messy diff :\

evantahler:
I think that the right thing to do is in `servers/websocket.js:166` allow both `null` and `undefined`.

evantahler:
Bump.  Need any help here?

submitteddenied:
I have updated it so that falsey values will prevent client side JS from being generated

evantahler:
Great! Thanks!

PR #612, opened 2015-03-06
pushing a number of resque logs deeper

gcoonrod:
Trying to reduce the verbosity of resque logging.

### DIFF
--- a/config/tasks.js
+++ b/config/tasks.js
@@ -2,18 +2,18 @@ exports.default = {
   tasks: function(api){
     return {
       // Should this node run a scheduler to promote delayed tasks?
-      scheduler: true,
+      scheduler: false,
       // what queues should the taskProcessors work?
       queues: ['*'],
-	  // Verbosity of task logging
-	  verbose: true,
+      // Verbosity of task logging
+      verbose: true,
       // how long to sleep between jobs / scheduler checks
       timeout: 5000,
       // at minimum, how many parallel taskProcessors should this node spawn?
       // (have number > 0 to enable, and < 1 to disable)
-      minTaskProcessors: 1,
+      minTaskProcessors: 0,
       // at maximum, how many parallel taskProcessors should this node spawn?
-      maxTaskProcessors: 3,
+      maxTaskProcessors: 0,
       // how often should we check the event loop to spawn more taskProcessors?
       checkTimeout: 500,
       // how many ms would constitue an event loop delay to halt taskProcessors spawning?
--- a/config/tasks.js
+++ b/config/tasks.js
@@ -2,16 +2,18 @@ exports.default = {
   tasks: function(api){
     return {
       // Should this node run a scheduler to promote delayed tasks?
-      scheduler: false,
+      scheduler: true,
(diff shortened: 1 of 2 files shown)


evantahler:
I'm personally of the opinion that starging/stopping workers should be of the `info` level, but lets here what others think

**comments welcome**

gcoonrod:
When using multiworkers i was seeing workers stopped and started every second. Which really cluttered up the logs. What is the rationale for having those events at the info level?

evantahler:
The rationale was in production the lowest log level you probably might use would be "info".  On production you are likely to be using a log aggression tool like Splunk, Sumologic, etc, and you probably want to know how many of your workers are running at any time.  You can then scan for the latest log of this type to see how many are up and running.

neilstuartcraig:
Info seems like the right choice to me too. It's certainly not higher than that. I guess you could make a case for debug but I'd expect to see it in info. 
For reference, you could check what apps like nginx etc do. Not saying they're becessarily right but could be validation.

gcoonrod:
Any thoughts on making the worker log levels controllable?

evantahler:
lets see what a PR for that would look like @gcoonrod! Might make sense depending on how much cruft that adds to the config file

gcoonrod:
@evantahler, I added an attribute to the tasks config and set some of the node-resque messages to use that attribute to determine their logging level. I only applied that option to the messages that have been bugging me. 

Are there other messages that this option could/should be applied to?

config/tasks.js
```diff
@@ -2,16 +2,18 @@ exports.default = {
   tasks: function(api){
     return {
       // Should this node run a scheduler to promote delayed tasks?
-      scheduler: false,
+      scheduler: true,
```
evantahler: Don't commit this

config/tasks.js
```diff
@@ -2,16 +2,18 @@ exports.default = {
...
       // what queues should the taskProcessors work?
       queues: ['*'],
+	  // Verbosity of task logging
+	  verbose: true,
       // how long to sleep between jobs / scheduler checks
       timeout: 5000,
       // at minimum, how many parallel taskProcessors should this node spawn?
       // (have number > 0 to enable, and < 1 to disable)
-      minTaskProcessors: 0,
+      minTaskProcessors: 1,
```
evantahler: and don't commit these.  we should start default projects with no workers

gcoonrod:
@evantahler Any further comments or suggestions?

config/tasks.js
```diff
@@ -5,6 +5,8 @@ exports.default = {
       scheduler: false,
       // what queues should the taskProcessors work?
       queues: ['*'],
+      // Verbosity of task logging
+      verbose: true,
```
evantahler: I don't think that I like having a log option outside of the logging config file.

Perhaps the answer is a hash with each event => log level?

```js
{start: "debug", poll: "trace"}
```

evantahler:
Lets merge it in for now, although I do expect some future changes.

evantahler:
Thanks!

PR #616, opened 2015-03-20
Clean up templates.

hippich:
Noticed that generated components have missing semicolons, so decided to update templates to include them, so at least eslint/jslint/jshint has less stuff to complain about.


evantahler:
Cool, thanks!

PR #619, opened 2015-03-26
Complex objects as ActionheroClient's webAction() parameters

samuelbrian:
When the ActionheroClient makes requests over HTTP, it constructs a string to pass as URL parameters to a GET request. As in:

    for (var i in params) {
        qs += i + '=' + params[i] + '&';
    }

This is troublesome for parameter objects such as:

    { a: { b: 3 } }

I've modified the webAction() to stringify and POST the parameter data if it is an object.

Although thinking now, the parameter argument should *always* be an object, making the original GET request code unnecessary if the new POST code is preferred.

### DIFF
--- a/client/actionheroClient.js
+++ b/client/actionheroClient.js
@@ -153,11 +153,11 @@ ActionheroClient.prototype.action = function(action, params, callback){
   }
 }
 
-ActionheroClient.prototype.actionWeb = function (params, callback) {
+ActionheroClient.prototype.actionWeb = function(params, callback) {
   var xmlhttp = new XMLHttpRequest();
   xmlhttp.onreadystatechange = function () {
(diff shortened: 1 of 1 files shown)


evantahler:
good ideas here! 
I'm happy to go all `POST` all the time for simplicity.  Please make that update, and I'll merge it in!

samuelbrian:
Done! Although I left in being able to override the HTTP method.

Also, I don't really know what is going on with the automated testing. I'm certain I didn't touch any of the code that caused the failed tests.

evantahler:
Looks like github was having issues when you added that lasts commit... re-running tests now...

evantahler:
ok, travis is happy now.
merging!

Thanks!

PR #625, opened 2015-04-04
new middleware syntax

evantahler:
New Middleware Syntax:

## Action Middleware:

Action middleware syntax is now:

```javascript
api.actions.addMiddleware(definition);
// I.E.
api.actions.addMiddleware({
  name: 'userId checker',
  global: false,
  priority: 1000,
  preProcessor: function(data, next){
    if(!data.params.userId){
      next(new Error('All actions require a userId') );
    }else{
      next(err);
    }
  }
  postProcessor: function(data, next){
    if(data.thing.stuff == false){
      data.toRender = false;
    }
    next(err);
  }
});
```

- action middleware can now define both a preProcessor and postProcessor as a tuple.
- action middleware by default is applied to no actions, unless you define `definition.global = true`
- actions now opt into actions with the `action.middleware` array where names of actions are supplied which are to be invoked by each action, ie:

```javascript
exports.randomNumber = {
  name: 'randomNumber',
  description: 'I am an API method which will generate a random number',
  outputExample: {
    randomNumber: 0.123
  },

  middleware: [ 'userId checker' ]
  
  run: function(api, data, next){
    data.response.randomNumber = Math.random();
    next(null);
  }

};
```
## Connection Middleware:

Connection middleware syntax is now:

```javascript
api.connections.addMiddleware(definition);
// I.E.
api.connections.addMiddleware({
  name: 'connection middleware',
  priority: 1000,
  create: function(connection){
    // do stuff
  },
  destroy: function(connection){
    // do stuff
  }
});
```

- the connection middleware is still async (it has no callback; nor does it return).  It will not block the connection being created.

## Chat Middleware (say)

```javascript
api.chatRoom.addMiddleware(definition);
// I.E.
api.connections.addMiddleware({
  name: 'connection middleware',
  priority: 1000,
  join: function(connection, room, callback){
    // do stuff
    callback(err);
  },
  leave: function(connection, room, callback){
    // do stuff
    callback(err);
  },
  say: function(connection, room, messagePayload, callback){
    // do stuff
    callback(err, messagePayload);
  }
});
```

### DIFF
--- a/bin/templates/action.js
+++ b/bin/templates/action.js
@@ -6,6 +6,7 @@ exports.action = {
   matchExtensionMimeType: false,
   version:                1.0,
   toDocument:             true,
+  middleware:             [],
 
   inputs: {},
 
--- a/config/api.js
+++ b/config/api.js
@@ -30,7 +30,7 @@ exports.default = {
       missingParamChecks: [null, '', undefined],
       // The default filetype to server when a user requests a directory
       directoryFileType : 'index.html',
-      // The default priority level given to preProcessors, postProcessors, createCallbacks, and destroyCallbacks
+      // The default priority level given to middleware of all types (action, connection, and say)
       defaultMiddlewarePriority : 100,
       // configuration for your actionhero project structure
       paths: {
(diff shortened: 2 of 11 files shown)


S3bb1:
Nice Feature! This looks great when it comes to authentication with the ah-passport plugin! Good Work @evantahler !

PR #630, opened 2015-04-07
add stickyParams configuration option

heavyk:
:) cheers

### DIFF
--- a/config/servers/websocket.js
+++ b/config/servers/websocket.js
@@ -46,6 +46,10 @@ exports.default = {
           // transport:        {},
           // queueSize:        Infinity,
         },
+
+        // Parameters you wish to have persisted between requests
+        // Common use cases will be an authentication token or a user id
+        stickyParams: ['token']
       }
     }
   }
--- a/servers/websocket.js
+++ b/servers/websocket.js
@@ -26,7 +26,8 @@ var initialize = function(api, options, next){
       'roomView',
       'detailsView',
       'say'
-    ]
+    ],
+    stickyParams: api.config.servers.websocket.stickyParams || []
   }
 
   var server = new api.genericServer(type, options, attributes);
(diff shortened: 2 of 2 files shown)


heavyk:
I wrote this fix kinda fast... I realized you may not want 'token' by default a stickyParam (though, I'd say that's probably the only default acceptable and fits my use case). also it's silly to do `stickyParams: api.config.servers.websocket.stickyParams || []` (the `|| []` bit) if it's already defined. I was going to make it a property of genericConnecion but I didn't want to read so much code to figure out how to do that. this would be nice to be able to set it to `true` for example on socket connections and have all params persist like it is now.

I also broke your test :/

heavyk:
last thing: we needed this fix kinda fast but later after work I can rebase against v11.0 branch and add the `true` ability to the `stickyParameters` option (to persist all params and consequently not break existing people's socket connections) - additionally, I think this should be better located on the genericConnection prototype. I can think of other use-cases where it may be nice to decide which params to persist. so, yeah if you'd like me to do that and feel that's the direction to go, I'll do it, I made this PR just cause I wanted to contribute it back .. this fix was sufficient for our purposes. peace.

evantahler:
I'm sorry, but I think params need to be either all of nothing.  Allowing for *some* params to be sticky just seems confusing to me.

That said, you can probably to this in a middleware and ship it as a plugin!  You can store something like `connection.stuckParams = {}`.  In the middleware (pre action) you can populate any missing `connection.params` from `connection.stuckParams`, and then after the action, you can persist some select params to `stuckParams` for the next action.

heavyk:
fair enough. I did try with middleware, however there is a constraint that middleware is not fired for websocket connections (only web). there is literally no way to do this without modifying AH... (other than re-implement all of the websocket server myself)

this is why I had suggested that it become a part of genericServer.

heavyk:
which reminds me, is there a chance to get middleware on non-web connections?

evantahler:
absolutely! middleware does fire for all connection types.  You can report out `connection.type` and see what type of connection you have.  There are 3 types of middleware: action, connection, and chat  http://www.actionherojs.com/docs/core/middleware.html.  You probably want "action"

evantahler:
if you are having trouble with middleware firing with websocket connections, open up a new issue please.

heavyk:
I will have a look at it again tomorrow. I honestly can't really remember exactly what was the issue.

thank you for your time

Issue #632, opened 2015-04-12
inconsistient action return values


heavyk:
we had a new programmer using actionhero, and he ran into the same thing I did a long time ago, which is actionWeb and actionWebSocket are not consistent.. I have prepared a fiddle.. http://jsfiddle.net/0bckekto/10/

it stems from this block of code: https://github.com/evantahler/actionhero/blob/master/client/actionheroClient.js#L149-L153

essentially, he would call action immediately, so being disconnected still, it would use xhr, then after the websocket was connected, it would later use websocket, so error was no longer the first parameter to the callback. I might suggest this improvement (I'll make a PR in a sec):

```js
ActionheroClient.prototype.actionWebSocket = function(params, callback){
  this.send({event: 'action',params: params}, function(res) {
    if(res.error) {
      callback(res.error, res);
    }else{
      callback(null, res);
    }
  });
}

```

PR #633, opened 2015-04-12
consistent errors when using client#action

heavyk:
as suggested in #632

perhaps you may want to do it the other way around, and for xhr, instead `callback({error: err})`

your call.. cheers
[EDIT: fixed the commit msg]

### DIFF
--- a/client/actionheroClient.js
+++ b/client/actionheroClient.js
@@ -170,12 +170,18 @@ ActionheroClient.prototype.actionWeb = function(params, callback) {
   var url = this.options.url + this.options.apiPath + '?action=' + params.action;
   xmlhttp.open(method, url, true);
   xmlhttp.setRequestHeader('Content-Type', 'application/json');
-  xmlhttp.send(JSON.stringify(params));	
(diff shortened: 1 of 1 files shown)


evantahler:
The problem in your fiddle (mentioned in https://github.com/evantahler/actionhero/pull/632) is that you aren't waiting for the `client.connect()` callback (http://www.actionherojs.com/docs/servers/websocket.html). If wait for the callback properly, then you won't need to handle return values separately depending on if you are connected or not.

You only changed the return value style of action, but if you want to make this change, you should also update all the other commands, like `say`, `roomView`, etc.  Then, we can discuss the change.

evantahler:
Also, this breaks all the tests :(

heavyk:
Jaja... Actually I was thinking about it and I'd rater it all be on one
object and check the error property. I'll change it when I get home and
resend the request... Still haven't tried the middkeware thing. Today was a
long day at work

Cheers!

evantahler:
Closing due to lack of response.   Feel free to open again!

PR #638, opened 2015-04-20
Added (Literate) Coffeescript support for actions, config, initializers ...

hornos:
Added (Literate) Coffeescript support for actions, config, initializers and tasks.


evantahler:
Why would you require coffeescript in an otherwise non-coffeescript project?  Can these not be pre-compiled at boot?

hornos:
Literate coffeescript is probably the best lang for JS development. Coffeescript is compiled at boot by the coffee interpreter without any extra build step, it just works and no need to precompile. If you won't accept the request please allow some other method to define source extensions.

PhilWaldmann:
the `coffee-script` package should be an optional dependency. It won't make any sense to load this package for user which aren't using coffeescript.

but it's a good idea to offer that flexibility!

evantahler:
Sorry, I will not add support for any other language/compiler/transpiler into this project.   This is a *javascript* project, and only javascript will supported.  We will not support coffesscript, livescript, go-to-js, etc.  There are too many options and no clear winner at this time.

That said, you have a few options:
- You can change the behavior of `npm start` to first run the coffeescript compiler, and then start the server
  - You will loose development-mode live-reloading on changes of coffeescript files
- You can write an actionhero plugin which will load/include coffescript at boot

hornos:
Live reload works like a charm, you don't loose it. Can you give real reason why you against giving the option for more source extension? I do not want to include coffee just an option for more extensions besides js.

evantahler:
It's about scope and memory footprint of the application.  "users" of the framework can do whatever they want, but it's best to keep the core as small and lean as possible.   It's the same reason this project doesn't use anything like underscore.

evantahler:
Also, feel free to discuss your PR on the mailing list or chat!  If a lot of people want it, we will of course include it.

http://www.actionherojs.com/learn-more/

hornos:
I agree if you want to keep the core pure, however I do not see why a non-js source option (eg. by an ACTIONHERO environment var) violates the KISS principle of the core since it is up to the user if they want to use it or not. But without it you severly limit their ability to build upon the core and force js-only on everybody.

BoLaMN:
cheers @hornos this was handy to allow me to convert all of actionherojs over to actionherocoffee ;) id throw it on a fork but due to the apache 2.0 licensing trouble of adding change clauses to all the file headers and the other rigmarole needed for the apache 2.0 license.

evantahler:
I haven't heard much chatter about Coffeescript since ES6 came out.   Closing this issue

PR #640, opened 2015-04-21
Added fingerprint test-cases for websocket

dunse:
A better way is probably to remove the dependency of cookies while using websocket.
[16:16] <@`3rdEden> dunse: Clients cannot set headers because this doesn't work in browsers for all transports.
[16:17] <@`3rdEden> So if you want to authententicate I would suggest using querystrings instead
[16:17] <@`3rdEden> e.g. new Primus(https://example.domain.com?token=auth)


test/servers/websocket.js
```diff
@@ -401,28 +401,95 @@ describe('Server: Web Socket', function(){
...
+      cookieHeader = '';
+      connectClients(done);
+    });
+
+    before(function(done) {
+      // Override http.request to test fingerprint
+      var module = require('http');
+      oldRequest = module.request;
+      module.request = function(options, callback) {
+        options.headers.Cookie = cookieHeader;
```
evantahler: Fancy.

evantahler:
First off, thank you for solving how to test this! This has been bugging me for a while!

Seeing as how you mess with `http.request`, could you please move this into a new file, something like `/test/integrations/web_websocket`?  This helps us be *Really* sure that the changes to HTTP can be rolled back.

Regarding the way you want to auth via GET url... I'll open a new issue for it: https://github.com/evantahler/actionhero/issues/642

test/servers/websocket.js
```diff
@@ -401,28 +401,28 @@ describe('Server: Web Socket', function(){
...
   describe('disconnect', function(){
 
+    before(function(done) {
+      // Cleanup any stale connections
+      for(var id in api.connections.connections){
+        api.connections.connections[id].destroy();
+      }
+      done();
+    });
+
```
evantahler: What's the point of adding this `before` block now?

PR #643, opened 2015-04-21
add default params to routes

PhilWaldmann:
Hey,

currently there is no way to add a route (e.g `POST /old_login`) which will target a specific version of an action. The only way to achieve this is to add `?apiVersion=x` to the request, which is problematic if you have for example some legacy mobile apps which can't be updated.

I've added the possibility to define default params for every route:
```js
post:{
  { path: '/old_login', action: 'login', params:{ apiVersion: '1' } }
}
```

this will also work for other params beside `apiVersion`.

Tests are included!

### DIFF
--- a/initializers/routes.js
+++ b/initializers/routes.js
@@ -17,10 +17,8 @@ module.exports = {
           var route = api.routes.routes[method][i];
           var match = api.routes.matchURL(pathParts, route.path);
           if(match.match === true){
-            if(route.params){
-              for(var param in route.params){
-                connection.params[param] = connection.params[param] || route.params[param];
-              }
+            if(route.apiVersion){
+              connection.params.apiVersion = connection.params.apiVersion || route.apiVersion;
             }
             
             for(var param in match.params){
@@ -80,8 +78,8 @@ module.exports = {
     }
 
 
-    api.routes.registerRoute = function(method, path, action, params) {
-      api.routes.routes[method].push({ path: path, action: action, params:params});
+    api.routes.registerRoute = function(method, path, action, apiVersion) {
+      api.routes.routes[method].push({ path: path, action: action, apiVersion:apiVersion});
     }
 
     // load in the routes file
@@ -104,10 +102,10 @@ module.exports = {
           if(method === 'all'){
             for(v in api.routes.verbs){
               verb = api.routes.verbs[v];
-              api.routes.registerRoute(verb, route.path, route.action, route.params);
+              api.routes.registerRoute(verb, route.path, route.action, route.apiVersion);
             }
           } else {
-            api.routes.registerRoute(method, route.path, route.action, route.params);
+            api.routes.registerRoute(method, route.path, route.action, route.apiVersion);
           }
           counter++;
         }
--- a/initializers/routes.js
+++ b/initializers/routes.js
@@ -17,6 +17,12 @@ module.exports = {
           var route = api.routes.routes[method][i];
           var match = api.routes.matchURL(pathParts, route.path);
           if(match.match === true){
+            if(route.params){
+              for(var param in route.params){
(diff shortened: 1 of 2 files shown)


evantahler:
what's wrong with:

```javascript
routes = {
  post: [
    {path: '/myAction', action: 'myAction'},
    {path: '/myAction/:apiVersion', action: 'myAction'}
  ]
}
```

PhilWaldmann:
if you have an action `foo` (`/api/foo`) which is at version 1, it's not possible to keep foo (version 1) via `/api/foo` but have a second version (`/api/foo/2`).
The original `/api/foo` will then automatically point to the [highest version](http://www.actionherojs.com/docs/core/actions.html#versions) of `foo`!

evantahler:
That's exactly the behavior intended.  `/api/foo` will always be the latest version, and you can access any older versions with `/api/foo/version`.    I take it you want the default version to be *not* the highest version then?

PhilWaldmann:
The default version could be the highest, but it should be able to target a specific version without specifying the version number in the url.

For example you have a legacy application (without an upgrade path) and the url it uses is `/api/foo`.
Adding a new version for action `foo` (with different input or output variables) will break that old application.

With this pull request it's possible to define the following:
```js
routes = {
  get: [
    {path: '/foo', action: 'foo', params:{apiVersion:1}},
    {path: '/foo/:apiVersion', action: 'foo'}
  ]
}
```

`api/foo` will point to version 1
`api/foo/2` will point to version 2

evantahler:
Hmm... I don't think that I like this because it feels like duplication of the `default` variable options that actions already have.  The only problem is that apiVersion is how you pick which action to run, so this would be the only case which can't be handled by the defaults in the action.

So perhaps the api gets a little simpler, like:  `{path: '/foo/:apiVersion', action: 'foo', apiVersion: 123}`.  I don't think I would want to use routes to modify any other params

dangerdespain:
Is parameter injection at the route level against best practices? I could see this being useful for other things. Being able to create a reusable getChildren action with scope-limiting done through the route definitions would be pretty useful in my project and would make for faster prototyping. Don't want to expose (or filter) all the combinations of parent/children relationships with a { path: '/:parent/:id/:child', action: 'getAllChildren' } route but want to quickly add an endpoint for GET /projects/:id/campaigns route?

It'd be easy with 
{ path: '/projects/:id/campaigns', action: 'getChildren', params : { parent : 'projects', child : 'campaigns' } }

My project is loaded with a couple dozen actions like campaignGetTasks, projectGetCampaigns, campaignGetNotifications and they could all be distilled into the same action with parameter injection. Seems appealing to me.

PhilWaldmann:
Evan is right, adding `params` is a kind of duplication.
Routes are for web only, so it won't work for any other server-type anyways.

But adding the `apiVersion` config option to a route will totally make sense!

To solve @dangerdespain problem it would be nice to have some kind of action alias with predefined params. This alias actions could be defined like a normal action but without a run function.

dangerdespain:
Good point, @PhilWaldmann! I haven't used WS/sockets in AH yet so thanks for chiming in about that. I do like Evan's approach of using apiVersion as its own parameter. This would provide a good option for managing versions for people who didn't integrate API version selection into their clients (note to self... integrate API version selection into my project!)

evantahler:
Lets worry about actions calling other actions in another PR :D (this sounds like middleware to me).

I'll be happy with this PR if we only care about a default `apiVersion`.

Also be sure to have a matching PR with changes to the documentation!

PhilWaldmann:
I'll change this pull request tomorrow.

evantahler:
awesome, thanks!

evantahler:
docs related to this change added here https://github.com/evantahler/actionhero/commit/ca6717cd096261e07fc1d6dd51c271c13a779c19

PR #646, opened 2015-04-27
New say chat middleware

dunse:
Added new direction for the chat middleware called 'onSayReceive'.
This middleware gets executed just before message is published (only once).
Message Payload is the same as for the 'say' middleware.
Left 'say' as is, just to maintain backward compatibility.
Added test-cases for both 'say' and 'onSayReceive'.


dunse:
Could you please re-run the Travis build? I think it had some issues when it built this pull request..

evantahler:
Looks like there is something wrong with Travis... lets give it a few hours and try again.

evantahler:
looks like this got out of date... can you merge master and try again?

Thanks!

dunse:
It failed for:
> 1) Core: Cache cache.load without changing the expireTime will re-apply the redis expire:
> Uncaught AssertionError: expected 'Error: Object expired' to be 'Error: Object not found'

Seems to be some timing issue.. Could you please re-run the build?

evantahler:
Tests are good now.

evantahler:
Cool.  

Now, I moved the docs into a sub folder of the main project (rather than another branch), so all pull requests like this can also update the docs :D

Be sure to change 
- https://github.com/evantahler/actionhero/blob/master/docs/docs/core/middleware.md
- https://github.com/evantahler/actionhero/blob/master/docs/docs/core/chat.md

evantahler:
bump! Any progress on the docs?

dunse:
I was looking at updating the docs fetched from master branch, but they do not seem to match what's published on the site. The branch gh-pages seems to be what is still in use. Please advice how you would like to proceed.

evantahler:
The way things work now is that the docs in MASTER (or your branch) are expected to be ahead of what's on live (gh-pages).  When we push to npm, we also sync the docs branch.  This allows us to write new docs for new features, but not confuse folks who are using the currently released version.  

https://github.com/evantahler/actionhero/blob/master/package.json#L65-L66

Issue #644, opened 2015-04-22
--node options with --daemon key


rootStar-lock:
Is there any any cli options to send params like: --expose-gc, --nouse-idle-notification, to node when i'm start with --daemon key?

evantahler:
no, you would need to send these manually:  `node --expose-gc --nouse-idle-notification ./node_modules/.bin/actionHero start --daemon `

rootStar-lock:
Sorry but this don't work for me.
In fork I just add ```data.response.gc_exposed = global.gc ? true : false;``` to __actions/status.js__ 
and run `$ node --expose-gc --nouse-idle-notification bin/actionhero start --daemon` gc_exposed was __false__ all the time.
iojs v1.8.1

PR #647, opened 2015-04-27
Added --node-x param to pass to node -x when starting in --daemon mode

rootStar-lock:
#644 

Examples
-------------
`$ node bin/actionhero start --node-debug --node-nouse-idle-notification` - will spawn daemonize process with __--debug__ and  __--nouse-idle-notification__ switch.
`$ node bin/actionhero startCluster --node-expose-gc --daemon` - will spawn cluster with __--expose-gc__ for each fork.
etc.


actions/status.js
```diff
@@ -42,9 +42,10 @@ exports.status = {
         data.response.stats             = stats;
         data.response.queues            = details.queues;
         data.response.workers           = details.workers;
-        
+        data.response.gc_exposed        = global.gc ? true : false;
```
evantahler: please don't check this in.

rootStar-lock: Sorry

bin/actionhero
```diff
@@ -175,17 +175,18 @@ binary.isDaemon = (binary.argv.isDaemon === 'true');
 if(binary.actions[binary.mainAction]){
   if(binary.argv.daemon){
     var args = process.argv.splice(2);
+    var nodeArgs = [];
+    var workArgs = [path.normalize(binary.paths.actionheroRoot + '/bin/actionhero'), binary.mainAction, '--isDaemon=true'];
```
evantahler: why are you hard-coding the daemon options?  This should be an option :D

rootStar-lock: OMG! A see I foget some more important that hard-coded --isDaemon. Otherwise, is [if(binary.argv.daemon){](https://github.com/evantahler/actionhero/blob/master/bin/actionhero#L176) is not guarantee that we must spawn child with --isDaemon=true?

evantahler: The way it works is that if you indend to boot actionhero (cluser or otherwise) it means that we need to spawn a child process that isn't bound to your current shell.  So to do this, the first process (the one you invoke directly) has the `dameon` flag passed too it.  It then forks, and and the new process has the `isDameon` flag in its args.   The main process can then quit once it knows that a child has been booted with the isDameon ags.

rootStar-lock:
Tested with 
1) `node bin/actionhero startCluster --daemon --node-expose-gc --workers=4 --node-debug --title HelloCluster --node-harmony --workerTitlePrefix Worker`
2) `node bin/actionhero start --node-expose-gc --node-debug --title HelloCluster --node-harmony --daemon`

bin/actionhero
```diff
@@ -175,17 +175,21 @@ binary.isDaemon = (binary.argv.isDaemon === 'true');
...
+
+    var nodeArgs = [];
+    var workArgs = [path.normalize(binary.paths.actionheroRoot + '/bin/actionhero')].concat(args);
+
     for(var i in args){
-      if(args[i].indexOf('--daemon') >= 0){
-        args.splice(i, 1);
-        break;
+      if(args[i].indexOf('--node') >= 0){
+        nodeArgs.push('-' + args[i].slice(6));
```
evantahler: the `slice(6)` implies that `--node` is the first argument, which might not be the case.  For example in ubuntu, the `nodejs` executable might be used.   Also, any symlinks (perhaps iojs) might break it.

bin/actionhero
```diff
@@ -175,17 +175,21 @@ binary.isDaemon = (binary.argv.isDaemon === 'true');
...
+        workArgs.splice(workArgs.indexOf(args[i]), 1);
       }
     }
-    args.push('--isDaemon=true')
-    var command = path.normalize(binary.paths.actionheroRoot + '/bin/actionhero');
-    var child = spawn(command, args, {detached: true, cwd: process.cwd(), env: process.env, stdio: 'ignore' });
+
+    workArgs[workArgs.indexOf('--daemon')] = '--isDaemon';
```
evantahler: This is weird to me.  If `--daemon ` is not present in `workArgs`, then the index will be a negative number, lets say `-5`.  So if you assign a value to a negative position in some js environments, strange things happen:

```
> a = []
[]
> a[-5] = '--isDaemon'
'--isDaemon'
> a
[ '-5': '--isDaemon' ]
```

bin/actionhero
rootStar-lock: No --node that is --node-x param where x is debug, expose-gc etc. Thats params splits and normilize to --x format. The are no connection between executable and this params. Executable to spawn child process always get from process.argv[0] which I assume the correct way.

bin/actionhero
rootStar-lock: Ok let imagine that line of code(`workArgs[workArgs.indexOf('--daemon')] = '--isDaemon';`) will be executed when we have no `--daemon` param. Then we have -1 constantly. But there are no chance to be with [if(binary.argv.daemon){](https://github.com/evantahler/actionhero/blob/master/bin/actionhero#L176) 

p.s I can rewrote base code use [process.execArgv](https://nodejs.org/api/process.html#process_process_execargv) and it will go with minimal changes but it block us to spawn __AH__ simple like `$ actionhero startCluster --daemon --node-debug --node-harmony`.

evantahler:
Any updates?

rootStar-lock:
Ok i minimize own code. Now we can use node arguments like
`$ node --nouse-idle-notification --expose-gc bin/actionhero start(startCluster) --daemon`

evantahler:
Any update here?

evantahler:
Closing due to lack of interest by poster.

PR #648, opened 2015-04-27
filter specific parameters in specific actions in log

levansuper:
added filterInLog parameter to  action.inputs that will filter the specific parameter of the specific action in log. usage : {inputs : { param : { filterInLog : true }}}.

### DIFF
--- a/initializers/actionProcessor.js
+++ b/initializers/actionProcessor.js
@@ -98,7 +98,8 @@ module.exports = {
       
       var filteredParams = {}
       for(var i in self.params){
-        if(api.config.general.filteredParams && api.config.general.filteredParams.indexOf(i) >= 0){
+        if((api.config.general.filteredParams && api.config.general.filteredParams.indexOf(i)) >= 0
+            || (self.actionTemplate.inputs[i] && self.actionTemplate.inputs[i].filterInLog===true)){
           filteredParams[i] = '[FILTERED]';
         }else{
           filteredParams[i] = self.params[i];
--- a/initializers/actionProcessor.js
+++ b/initializers/actionProcessor.js
@@ -99,7 +99,10 @@ module.exports = {
       var filteredParams = {}
       for(var i in self.params){
         if((api.config.general.filteredParams && api.config.general.filteredParams.indexOf(i)) >= 0
-            || (self.actionTemplate.inputs[i] && self.actionTemplate.inputs[i].filterInLog===true)){
+            || (self.actionTemplate 
+                && self.actionTemplate.inputs
+                && self.actionTemplate.inputs[i] 
+                && self.actionTemplate.inputs[i].filterInLog===true)){
           filteredParams[i] = '[FILTERED]';
         }else{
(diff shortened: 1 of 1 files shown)


evantahler:
Can you share your use case here?

evantahler:
This also fails the test suite.

levansuper:
im not sure about the tests and why they fail. i just added one line and you can check it.

but about use case:

i have this action that gets the byte code of the file and saves is. so i definitely dont want to have 5mb files byte code in the log.  
{
    "name": "FilesSaveFile",
    "description": 'saves file',
    "inputs": {
        "file": {
            "filterInLog" : true,
            "required": false
        }
    },
    "blockedConnectionTypes": [],
    "outputExample": {},
    "run": function(api, connection, next) {
        var file = connection.params.file1;      
        var date = new Date();
        //rootDir is set in one of my initializers
        fs.writeFile(api.rootDir) + date.getTime(), new Buffer(file));
        connection.response.result = {};
        next(null);
    }
}

and i have another input named "file" in another action where file really means "file name".
I know the naming in not perfect but that's the interface and i cant change it now because a lot of apps use it.  so i need to log in the second case and don't want to log in the firs case

evantahler:
haha, we all know the joys of needing to shim an exiting API.

I'm always trying to avoid adding more options to the action API/structure, as it is already so verbose & complex.  I understand your usecase too.  We already have the ability to `filter` a param entirely (made for credit card numbers and such), but I guess that won't work for you if you also want to show the filename in that other action.

 Perhaps we should generically change the logger.  How would you feel about: 
1. adding an option for `maxLogDataLength` or similar.  We would not log any param value longer than, say 30 chars, and would just show the first 30 letters.  We could also do this via bytes.
2. type checking.  I imagine somewhere you convert `data.params.file` into a buffer?  We should probably not log anything that isn't a `string` or `number` maybe?
3. remove it from data.params.   Also, you can just `delete data.params.file` at the end of your action, and it won't be logged.

levansuper:
The first and second cases are very similar and can be accomplished but i think the flexibility is lost if there is no config to allow logging, because in some cases i might be willing to log them.
the third is just i need but not having config field in the input and deleting the parameter in "run" method will make it a little confusing for people who doesn't know Actionhero and  have to work on the existing project.
that's in my humble opinion :)

evantahler:
Yeah, I don't like 3 either, but you can use it right now if you are stuck.

I would personally vote for 2 in this case.  Do you convert the params to another type from string?

levansuper:
i'm directly sending an array of bytes and  putting into the buffer so i'm not using the string at all

evantahler:
Actually, lets do this https://github.com/evantahler/actionhero/pull/650

I'll implement the string length limit, but move `api.actionProcessor.prototype.logAction` to its own method.  Now you can just overwrite it :D  I think your use case is unique, so I don't want to make any changes to core yet, but I do want to allow you to "hack" the tool to get you moving.

evantahler:
Merged!

PR #652, opened 2015-05-04
exception handler needs action name

mstdokumaci:
there is this line in action exception handler: "simpleName = data.connection.action;"


evantahler:
The point of the new actionProcessor is to *not* need to ever modify the connection (to enable task parallelism).  I'll fix this in the in the exceptionHandler.

evantahler:
https://github.com/evantahler/actionhero/commit/6d818428ca9d82bdb4bc5361d437585a260df608

evantahler:
good catch!

PR #653, opened 2015-05-10
stats memory leak

BoLaMN:
live reload caused writeIncrements to create new timers before clearing the old timer

### DIFF
--- a/initializers/stats.js
+++ b/initializers/stats.js
@@ -34,14 +34,14 @@ module.exports = {
             multi.exec(function(){
               started--;
               if(started === 0){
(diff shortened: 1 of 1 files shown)


evantahler:
You probably also should `clearTimeout` before each invocation as well.

BoLaMN:
It does a clear at the start of the writeIncrements function but I can remove that and place a clearTimeout before each setTimeout if you wish. 

The current commit has fixed the issue thou.

Up to you.

evantahler:
ahh! Oops.

Thanks!

PR #655, opened 2015-05-12
add support for removing stuck old workers (resque v0.13)

S3bb1:
- added config for disabling/enabling the removeStuckWorkers function
- added logmessages when stuck workers are removed
- increased package.json for node-resque

### DIFF
--- a/config/tasks.js
+++ b/config/tasks.js
@@ -21,7 +21,10 @@ exports.default = {
       // When we kill off a taskProcessor, should we disonnect that local redis connection?
       toDisconnectProcessors: true,
       // What redis server should we connect to for tasks / delayed jobs?
-      redis: api.config.redis
+      redis: api.config.redis,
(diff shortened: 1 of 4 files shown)


S3bb1:
hm travis don't loves me ... but i think my changes does not affect the failed test

evantahler:
It's fine to bump the `node-resque` version, but I don't even want to allow this as a boot option (It is way to dangerous).    I think that we should note in the docs that this is a thing you can do, but it's up to the user to implement this in their own project.

evantahler:
This issue will be solved by this PR https://github.com/evantahler/actionhero/issues/654#issuecomment-101665471

evantahler:
Let me know when you make those changes!

S3bb1:
I've moved the logic to the docs and reused the text from node-resque, Hope this is okay for you @evantahler

evantahler:
Merged!

PR #658, opened 2015-05-20
first shot at utilising swagger docs

### LINKED ISSUE #44: refactor actionsView/self-documentation into a known format
Perhaps Swagger?
https://github.com/wordnik/swagger-core/wiki/API-Declaration

BoLaMN:
thought id integrate the swagger ui in a more built in way other than throwing it in the public folder plus this way we can utilize updates from npm install aswell

the swagger json is pretty basic atm due to the lack of information i can get from the action views, the only information i can really get is if the parameters required... (maybe something to look at?)

still need to write some tests to cover this new code but thought id get something up for review. 

i wrote it all in coffeescript but the initializer looks pretty bad compiled so ill work on cleaning that up for native javascript. 

resolves #44

cheers


actions/swaggerDocumentation.js
```diff
@@ -0,0 +1,13 @@
+exports.swaggerDocumentation = {
```
evantahler: This file should replace the existing `documentation` action

actions/swaggerDocumentation.js
```diff
@@ -0,0 +1,13 @@
...
+  name: 'swaggerDocumentation',
+  description: 'Returns Swagger Documentation',
+  outputExample: 'Swagger JSON',
+
+  run: function(api, data, next) {
+  	if (api.config.servers.web && api.config.servers.web.enabled &&  api.config.servers.web.swaggerEnabled) {
```
evantahler: The JSON can still be sent to non-HTTP clients

config/servers/web.js
```diff
@@ -19,6 +19,12 @@ exports.default = {
           'Access-Control-Allow-Methods': 'HEAD, GET, POST, PUT, PATCH, DELETE, OPTIONS, TRACE',
           'Access-Control-Allow-Headers': 'Content-Type',
         },
+        // enable or disable the swagger endpoint 
+        swaggerEnabled: true,
```
evantahler: `documentation` is always enabled, so remove this option.  Disabling it is simply removing the action

config/servers/web.js
```diff
@@ -19,6 +19,12 @@ exports.default = {
...
+        // rename or change the swagger endpoint http://hose:port/{urlPathForSwagger} || http://hose:port/{urlPathForSwagger}/
+        urlPathForSwagger: 'swagger',
```
evantahler: not needed, the action's name covers this (or folks can add a route)

config/servers/web.js
```diff
@@ -19,6 +19,12 @@ exports.default = {
...
+        // the action hero action name for the swagger json 
+        swaggerAction: 'swaggerDocumentation',
```
evantahler: not needed, the action just "exists"

evantahler:
Good start!

- As you note, the initializer is a bit of a mess.  I'm looking forward to the JS-native version
- I don't think that anything for the swagger UI should be included in the web server.  Why can't we include a static HTML file?  We can include a grunt task to generate the static HTML.

initializers/swagger.js
```diff
@@ -0,0 +1,171 @@
...
+    var actionUrl, bindIp, serverPort;
+
+    if (config.servers.web && config.servers.web.enabled) {
+      actionUrl = config.servers.web.urlPathForActions;
+      bindIp = config.servers.web.bindIP || 'http://localhost/';
+      serverPort = config.servers.web.port || '';
+    };
+
+    api.swagger = {
+      actionToSwagger: {
```
evantahler: what does this hash-key mean?

package.json
```diff
@@ -40,6 +40,7 @@
     "optimist": "0.6.1",
     "primus": "3.0.2",
     "redis": "0.12.1",
+    "swagger-ui": "^2.1.5-M2",
```
evantahler: This needs to be a devDependancy.  This is not required to run actionhero.

This is one of the main reasons to move the generation of the static HTML to a grunt task

servers/web.js
```diff
@@ -6,6 +6,7 @@ var formidable          = require('formidable');
 var browser_fingerprint = require('browser_fingerprint');
```
evantahler: The web server should not be modified at all for this PR.  Slowing down *every* web requests with more conditionals/logic for documentation is unacceptable.

test/servers/web.js
```diff
@@ -636,14 +636,16 @@ describe('Server: Web', function(){
       it('can ask for nested URL files', function(done){
         request.get(url + '/a/b/c/simple.html', function(err, response){
           response.statusCode.should.equal(200);
-          response.body.should.equal('<h1>ActionHero</h1>\\nI am a flat file being served to you via the API from ./public/simple.html<br />');
```
evantahler: The file is here! https://github.com/evantahler/actionhero/blob/master/public/simple.html

test/servers/web.js
```diff
@@ -636,14 +636,16 @@ describe('Server: Web', function(){
       it('can ask for nested URL files', function(done){
```
evantahler: This file shouldn't change.  

I *do* expect this test to change drastically https://github.com/evantahler/actionhero/blob/master/test/actions/showDocumentation.js

initializers/swagger.js
BoLaMN: swaggers json structure needed a host: 'host:port' object and the only thing i could see was the bindIp and port but due to https://github.com/evantahler/actionhero/blob/master/test/core/actionCluster.js i had to check to see if it the config values were avaliable or not. 

is there anywhere else that specifies host:port?

test/servers/web.js
BoLaMN: but its not /a/b/c/simple.html its /public/simple.html or am i missing something.. it is pretty late :/

evantahler:
Looks like the `responseCode` bit doesn't work:

screen shot 2015-05-20 at 2 46 55 pm

evantahler:
actually, the requestURL isn't being built properly on localhost: 
screen shot 2015-05-20 at 2 48 15 pm

package.json
BoLaMN: the deps for swagger-ui are pretty big that's why i sought after the web server serving the files from the node_modules folder instead of throwing them into the public folder which made it a bit more flexible for the end user instead of them having to keep on editing the index.html if they move the folder around.

initializers/swagger.js
evantahler: Lines 9-12 are OK, I'm asking about the key `actionToSwagger`

test/servers/web.js
evantahler: This test changes the routes, that's the point: https://github.com/evantahler/actionhero/blob/master/test/servers/web.js#L605-L608

BoLaMN:
just commit the fix for the api:8080 issue, nearly midnight here so missed it by accident.

package.json
evantahler: The large collection of deps is exactly why I don't want to require/include them in the project!   Making *everyone* who uses actionhero use swagger if they don't want to is a deal-breaker.  Making the documentation API "speak" swagger is OK (because the JSON format we picked is arbitrary), but the UI is huge, and this is an API sever.

If you can add the UI without requiring this package, that's great.  Otherwise, you should make a new plugin `ah-swagger-ui` or something like that that folks can opt-into.  Even without a UI people can just visit `http://petstore.swagger.io/` and point it at localhost and use the UI eternally.

evantahler:
All good! just commenting as a first pass.

BoLaMN:
made some of the required changes and removed the old documentation code ill continue on tomorrow as i need to get some rest.

evantahler:
awesome! Thanks so much for doing this! https://github.com/evantahler/actionhero/issues/44 is a 3-year old issue.

Ping me again when you are ready for the next round of reviews!

BoLaMN:
@evantahler been busy porting a backbone app over to angular so i hadnt had time to finish this PR off.

I cleaned up the initialiser and recreated the index.html and baked stats, chat and docs all together with angular 

i'll cleanup the app.js or i can add the seperate .js files and sass files if you want and add a grunt task or i can remove it all and create a seperate plugin as i wasnt sure how else to include swagger-ui without having to include all 37 dep files from the original swagger-ui

let me know, cheers

config/servers/web.js
```diff
@@ -19,6 +19,8 @@ exports.default = {
           'Access-Control-Allow-Methods': 'HEAD, GET, POST, PUT, PATCH, DELETE, OPTIONS, TRACE',
           'Access-Control-Allow-Headers': 'Content-Type',
         },
+        // rename or change the swagger endpoint http://hose:port/{urlPathForSwagger} || http://hose:port/{urlPathForSwagger}/
```
evantahler: `host` not `hose`

initializers/genericServer.js
```diff
@@ -92,6 +92,14 @@ module.exports = {
       actionProcessor.processAction();
     }
 
+    api.genericServer.prototype.processSwagger = function(connection){
```
evantahler: this method doesn't seem to be used anywhere?

evantahler:
good updates! 

Please cleanup the public folder as you want it to be and I'll take a look.  I see that you currently have removed chat.html and actionhero's css.  Why?

supamii:
FYI - I've been able to separate this (with tweaks) into a plugin for my own personal use - so far so good.

So like you suggested Evan, there's really no need for this to merge directly into AH itself.

Still, very nice work from Mr BoLaMN.  +props.

Also, let me know and I could possibly submit the npm plugin in some form.

evantahler:
of course you can make an npm plugin @supamii!  I'd talk to @BoLaMN about who wants to do it... this thread seems to have sadly gone cold :(

supamii:
@evantahler Ok, I'm currently customizing the base code for my own work but if i happen to cover enough bases i'll present a first shot at the plugin.  

There are surprisingly a lot of cases to handle for APIs that use only routes vs simple queries or action queries, etc etc etc.  Versioning for api routes is another beast all together.

BoLaMN:
hey guys, sorry ive been flat out on a work project so i haven't had time to continue this PR. my last PR the reason i removed the chat.html and actionhero.css is because i rewrote it all into an angular system that include stats, docs and chat to save including the swagger-ui dependencies

supamii:
Yeah I definitely saw that, pretty cool stuff you got going.  What I'm going to propose is a more passive approach to take in the swagger-ui dependencies and also a few fixes to handle routes better with model-schemas and even the option to include enums to toggle on.  All of this, will be made available as a plugin.  Really cool stuff you've done here, I'll keep you all posted.

evantahler:
I'm closing this thread.

- We now have a swagger plugin thanks to @supamii https://github.com/supamii/ah-swagger-plugin.  This new plugin looks like it was based on the work here. 
- Once some of the kinks get worked out in the plugin, we can talk about merging it into actionhero core and replacing the existing documentation system
- Converting all of actionhero's default pages to angular might be cool.  If you want to discuss that, please open a new pull request and we can talk about it in isolation.

guiguan:
Based on @supamii 's work, I changed to use a new swagger material ui. https://github.com/guiguan/ah-swagger-material-ui :)

l0oky:
wow looks slick, I am going to try that

PR #661, opened 2015-06-05
Fix unclear chat documentation for latest API

crrobinson14:
1. There are two roomStatus references in here, but the first one is only partially written.
1. The say() middleware callback requires a parameter (the message). If not passed into the callback, say() never sends anything.

### DIFF
--- a/docs/core/chat.md
+++ b/docs/core/chat.md
@@ -44,10 +44,7 @@ These methods are to be used within your server (perhaps an action or initialize
 
 ### api.chatRoom.roomStatus(room, callback)
 - callback returns (error, details); details is a hash containing room information
-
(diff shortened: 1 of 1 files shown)


docs/core/chat.md
```diff
@@ -44,10 +44,7 @@ These methods are to be used within your server (perhaps an action or initialize
 
 ### api.chatRoom.roomStatus(room, callback)
 - callback returns (error, details); details is a hash containing room information
-
-### api.chatRoom.roomStatus(room, callback)
```
evantahler: Why did you remove this method?  It's still there! https://github.com/evantahler/actionhero/blob/master/initializers/chatRoom.js

evantahler:
Nice.  Thanks!

PR #692, opened 2015-10-21
fix: fixes the error when using babel-node under strict mode

zhhz:
This pull request fixed the error when using babel-node to load the actionhero:  
    - package is a reserved keyword  => pkg
    - 0777 is not a valid octal number => 0o777

Sorry the diff might looks a lot of lines changed, but actually it is my vim editor changed the lf of the original file. I would like to suggest you consider using .editconfig.

### DIFF
--- a/config/redis.js
+++ b/config/redis.js
@@ -1,12 +1,12 @@
-exports.default = { 
+exports.default = {
   redis: function(api){
     var redisDetails = {
       // Which channel to use on redis pub/sub for RPC communication
       channel: 'actionhero',
-      // How long to wait for an RPC call before considering it a failure 
-      rpcTimeout: 5000, 
-      // which redis package should you ise?
-      package: 'fakeredis',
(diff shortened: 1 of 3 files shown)


evantahler:
I'm not going to merge this now, as a lot of the dependant packages use `opject.package`, which presumably would also break in node-babel.  I also feel we shouldn't make out programming less clear for transpiler.  

If enough people :+1: this issue, indicating high adoption for node-babel, we can re-visit.

zhhz:
Hi Evan,

Thanks for looking into this. I didn't realize there are issues with other dependent package. I agree with your decision. For now, I can live with that. I'm not sure how long it could be before node make ES6 as a standard, until then we are safe...

evantahler:
FYI this will be fixed in v13.0.0 of actionhero.

zhhz:
Cool. Thanks Evan!

PR #694, opened 2015-10-27
Slight JS Example Improvment

krishnaglick:
Since you're not doing a hard compare with null it's more javascripty to just check if `err` is truthy.


evantahler:
Thanks!

evantahler:
Whoops! In the future please don't work on the `gh-pages` branch, but rather the `docs` folder in master.

Replaced via https://github.com/evantahler/actionhero/commit/b130fe54a9a566213cc8e23237b94c38829582d3

krishnaglick:
Will do in the future, thanks!

PR #702, opened 2015-11-13
Change node-uuid to a relative dependency

crrobinson14:
node-uuid 1.4.6 just came out and looks like they nuked 1.4.4. Needs a relative semver rather than a hard-coded one.

Probably true for the others as well...


evantahler:
well that would explain things!

evantahler:
https://github.com/evantahler/actionhero/releases/tag/v12.2.2

PR #703, opened 2015-11-13
Update requirements to compatible-with

crrobinson14:
This changes the package requirements to compatible-with deps instead of hard-coded version numbers. Hopefully this will help prevent immediate failures if packages get unpublished in the future, like node-uuid 1.4.3 did.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,20 +30,20 @@
     "node": ">=0.10.0"
   },
   "dependencies": {
-    "async": "1.5.0",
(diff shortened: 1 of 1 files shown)


evantahler:
I'm totally down with this.   The only note is that folks would really need to ensure that they were using `npm shrinkwrap` when deploying... which I guess we all should be doing anyway...

evantahler:
Travis says that it is safe... so lets do it!

PR #704, opened 2015-11-15
move actionherojs.com site to Middleman and use Slate theme for docs

evantahler:
screen shot 2015-11-14 at 7 04 32 pm


evantahler:
This also includes a number of general documentation fixes and updates

PR #710, opened 2015-11-27
Fix broken documentation links

AngeloR:
I've updated the README so that links to the documentation point at the new Slate-based site.


evantahler:
Thanks!

PR #711, opened 2015-11-27
use the HAPI query parser to allow for nested param parsing

evantahler:
The hapi.js `qs` module is pretty great: https://github.com/hapijs/qs

It has configurable options for the multitude of options about how to parse the multitude of QS complex argument syntaxes. This allows arrays `?key[]=a&key[]=b` to work, dot notation `?key.a=a&key.b=b`, etc.  

We will default actionhero with none of the advanced features enabled, but the implementers of an api can easily add options for custom parsing via `api.config.servers.web.queryParseOptions`


evantahler:
closes:
- https://github.com/evantahler/actionhero/issues/709
- https://github.com/evantahler/actionhero/issues/649
- https://github.com/evantahler/actionhero/issues/314

evantahler:
Note: we cannot use the latest (v6) of `qs`, as they have moved to only ES6 (node >v1.0).  We will be locked to v5.x of QS so actionhero can still work on node v0.10 and up.

https://github.com/hapijs/qs/issues/124

joseym:
:+1:

Issue #719, opened 2015-12-16
Improvement/Question - Configurable task logging level


dangerdespain:
My console is starting to get noisy with all of the recurring tasks my latest project has running. Literally dozens of lines logged similar to this 
`2015-12-16 14:34:42 - info: worker: job success default workerId=1, class=schedulePosts, queue=default, result=null
` per minute.

I'd like to set up an option in api.config.tasks to set the logging level for tasks. Something like `api.config.tasks.loggingLevel = 'debug'` Would this be a welcome pull request?

evantahler:
Sure! 
Maybe we allow for log levels for each task event that you can customize, like:

```js
api.config.tasks.logLevels.tasks = {
  start: 'info',
  success: 'debug',
  failure: 'error',
  // ...
};

api.config.tasks.logLevels.scheduler = {};
```

This should also replace `api.config.tasks.verbose`

dangerdespain:
Love it! :+1:

dangerdespain:
Alright, details are in the pull request. As always, I'm open to input or tweaks!

PR #720, opened 2015-12-17
Implements task logging verbosity configuration options.

dangerdespain:
See the config file for the user-facing updates to task logging. The verbose option was left intact for scheduler logging, and is used to override the task verbosity. I did this to keep a single option around for quick debugging but I'm open to tweaking this!

This change is backwards-compatible and transparent to extisting configurations and includes documentation updates (Implements #719)

### DIFF
--- a/config/tasks.js
+++ b/config/tasks.js
@@ -5,13 +5,18 @@ exports.default = {
       scheduler: false,
       // what queues should the taskProcessors work?
       queues: ['*'],
-      // Verbosity of task scheduler
-      verbose: true,
       // Logging levels of tasks
       taskLogging : {
         start : 'info',
+        end : 'info',
         success: 'info',
         failure: 'error',
+        cleaning_worker : 'info',
+        poll : 'debug',
+        job : 'debug',
+        enqueue: 'debug',
+        reEnqueue: 'debug',
+        pause: 'debug',
       },
       // how long to sleep between jobs / scheduler checks
       timeout: 5000,
--- a/config/tasks.js
+++ b/config/tasks.js
@@ -5,8 +5,14 @@ exports.default = {
       scheduler: false,
       // what queues should the taskProcessors work?
       queues: ['*'],
-      // Verbosity of task logging
+      // Verbosity of task scheduler
       verbose: true,
+      // Logging levels of tasks
+      taskLogging : {
+        start : 'info',
+        success: 'info',
+        failure: 'error',
+      },
       // how long to sleep between jobs / scheduler checks
       timeout: 5000,
       // at minimum, how many parallel taskProcessors should this node spawn?
@@ -21,7 +27,7 @@ exports.default = {
       // When we kill off a taskProcessor, should we disconnect that local redis connection?
       toDisconnectProcessors: true,
       // What redis server should we connect to for tasks / delayed jobs?
-      redis: api.config.redis
+      redis: api.config.redis,
     }
   }
 }
--- a/config/tasks.js
+++ b/config/tasks.js
@@ -5,18 +5,28 @@ exports.default = {
       scheduler: false,
       // what queues should the taskProcessors work?
       queues: ['*'],
-      // Logging levels of tasks
-      taskLogging : {
-        start : 'info',
-        end : 'info',
-        success: 'info',
-        failure: 'error',
+      // Logging levels of task workers
+      workerLogging : {
+        failure   : 'error', // task failure
+        success   : 'info',  // task success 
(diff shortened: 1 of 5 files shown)


initializers/exceptions.js
```diff
@@ -98,7 +98,7 @@ module.exports = {
         simpleName = err.message;
       }
       var name = 'task:' + simpleName;
-      api.exceptionHandlers.report(err, 'task', name, {task: task, queue: queue, workerId: workerId}, 'error');
+      api.exceptionHandlers.report(err, 'task', name, {task: task, queue: queue, workerId: workerId}, api.config.tasks.verbose ? 'info' : api.config.tasks.taskLogging.success);
```
evantahler: This should be `failure`?

dangerdespain: Yes, good catch! Copypasta.. Fix inbound.

site/source/includes/docs/core/tasks.md
```diff
@@ -55,6 +55,14 @@ exports.default = {
       scheduler: false,
       // what queues should the TaskProcessors work?
       queues: ['*'],
+      // Verbosity of task scheduler (overrides the below task logging levels when set to true)
+      verbose: true,
+      // Logging levels of tasks
+      taskLogging : {
+        start : 'info',
+        success: 'info',
+        failure: 'error',
```
evantahler: I think you need options for **all** possible emitters: `poll`, `reEnqueue`, etc.

site/source/includes/docs/core/tasks.md
```diff
@@ -55,6 +55,14 @@ exports.default = {
       scheduler: false,
       // what queues should the TaskProcessors work?
       queues: ['*'],
+      // Verbosity of task scheduler (overrides the below task logging levels when set to true)
+      verbose: true,
```
evantahler: Why does the scheduler have a separate pattern?  It should match the worker and have a hash.

site/source/includes/docs/core/tasks.md
dangerdespain: I was keeping things simple and leaving the debug-level stuff alone. We can take it to that level of detail though, if you'd like.

evantahler: I would! it's more about having different patters for some emitters and not others.

dangerdespain:
So should the verbose option be stripped altogether, then? I like the idea of having a single override for debugging, but perhaps it's just too confusing.

evantahler:
I agree it should be removed

dangerdespain:
Alright.. you may still not like having all of these options under the taskLogging object. If that's the case, I'd like more specific direction on how you'd like the options split between taskLogging and schedulerLogging and would be happy to make those changes.

evantahler:
This is great! 
I think for the scheduler, you just make another hash, schedulerLogging, which has its events as well.  Yes, this makes the config file pretty large, but now we have *all* the options exposed to user-space, which is nice.

config/tasks.js
```diff
@@ -21,7 +32,7 @@ exports.default = {
       // When we kill off a taskProcessor, should we disconnect that local redis connection?
       toDisconnectProcessors: true,
       // What redis server should we connect to for tasks / delayed jobs?
-      redis: api.config.redis
+      redis: api.config.redis,
```
l0oky: Is that extra comma intentionally left there? `,` :smiley:

config/tasks.js
```diff
@@ -5,8 +5,19 @@ exports.default = {
...
+        start : 'info',
+        end : 'info',
+        success: 'info',
+        failure: 'error',
+        cleaning_worker : 'info',
+        poll : 'debug',
+        job : 'debug',
+        enqueue: 'debug',
+        reEnqueue: 'debug',
+        pause: 'debug',
```
l0oky: Trailing comma

evantahler:
Wait... why are trailing commas so bad?

neilstuartcraig:
It's invalid JSON, strictly speaking.
That graphic is awesome :-)

dangerdespain:
Haha you guys crack me up.

I'm coming down from a weekend in the mountains, should have this PR wrapped up tomorrow (:

dangerdespain:
Alright, my apologies for the delay. This ought to do the trick.The failing build seems to be unrelated timeouts.

Please be extra verbose if further tweaks are needed ;)

evantahler:
Yeah, sometimes travis times out.  I re-ran them and it's all green!

Thanks!

PR #728, opened 2015-12-29
enable `matchTrailingPathParts` in router

evantahler:
Enables a new option, `matchTrailingPathParts` which you can enable in the router to match combination paths: 

```javascript
post: [
  // yes match: site.com/api/123
  // no match: site.com/api/123/admin
  { path: '/login/:userId(/.*/)', action: 'login' }
],

post: [
  // yes match: site.com/api/123
  // yes match: site.com/api/123/admin
  { path: '/login/:userId(/.*/)', action: 'login', matchTrailingPathParts: true }
],
```

This also enables "catch all" routes, like: 

```javascript
get: [
  { path: ':path(/.*/)', action: 'catchAll', matchTrailingPathParts: true }
],
```

If you have a route with multiple variables defined and `matchTrailingPathParts` is true, only the final segment will match the trailing sections:

```javascript
get: [
  // the route site.com/users/123/should/do/a/thing would become {userId: 123, path: '/should/do/a/thing'}
  { path: '/users/:userId/:path(/.*/)', action: 'catchAll', matchTrailingPathParts: true }
],
```

### DIFF
--- a/initializers/routes.js
+++ b/initializers/routes.js
@@ -4,7 +4,6 @@ module.exports = {
 
     api.routes = {};
     api.routes.routes = {};
-
     api.routes.verbs = ['get', 'post', 'put', 'patch', 'delete'];
 
     ////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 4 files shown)


l0oky:
Looks cool. Getting closer to RESTful API develompent I guess.

evantahler:
It's actually further form RESTful development from what I understand: Assuming the object you are modifying can be manipulated in an obtuse (meaning more than one path is valid) way, is what REST explicitly tries to avoid.  

For example, `PUT users/123` would get a request to modify user 123 and `PUT users/123/house` would be a request to modify that user's house.  Explicitly requiring those to map to distinct actions enforces the separation of concerns.  Allowing both of those routes to hit the same action, and likely the same object, breaks the convention.

PR #736, opened 2016-01-09
v13.0.0

evantahler:
# `v13.0.0`

A number of changes (below) have updated the capabilities of the actionhero binary.  The new list (and help.txt) is:
```
actionhero - a node.js API framework for both tcp sockets, web sockets, and http clients.

Binary options:
* help (default)
* start
* startCluster
* generate
* generateAction
* generateTask
* generateInitializer
* generateServer
* actions
* enqueueTask
* console
* link

Descriptions:

* actionhero help
  will display this document

* actionhero start --config=[/path/to/config.js] --title=[processTitle]  --daemon
  will start a template actionhero server
  this is the respondent to "npm start"
  [config] (optional) path to config.js, defaults to "process.cwd() + '/' + config.js". You can also use ENV[ACTIONHERO_CONFIG].
  [title] (optional) process title to use for actionhero-s ID, ps, log, and pidFile defaults. Must be unique for each member of the cluster.  You can also use ENV[ACTIONHERO_TITLE]. Process renaming does not work on OSX/Windows
  [daemon] (optional) to fork and run as a new background process defaults to false

* actionhero startCluster --workers=[numWorkers]  --daemon
  will launch a actionhero cluster (using node-s cluster module)
  [workers] (optional) number of workers (defaults to # CPUs - 2)
  [daemon] (optional) to fork and run as a new background process defaults to false

* actionhero generate
  will prepare an empty directory with a template actionhero project

* actionhero generateAction --name=[name] --description=[description] --inputsRequired=[inputsRequired] --inputsOptional=[inputsOptional]
  will generate a new action in "actions"
  [name] (required)
  [description] (required) should be wrapped in quotes if it contains spaces

* actionhero generateTask --name=[name] --description=[description] --scope=[scope] --frequency=[frequency]
  will generate a new task in "tasks"
  [name] (required)
  [description] (required) should be wrapped in quotes if it contains spaces
  [scope] (optional) can be "any" or "all"
  [frequency] (optional)

* actionhero generateInitializer --name=[name]
  will generate a new initializer in "initializers"
  [name] (required)

* actionhero generateServer --name=[name]
  will generate a new server in "servers"
  [name] (required)

* actionhero actions
  will list all actions in this server to stdout

* actionhero enqueueTask --name=[taskName] --args=[JSON-formatted args]
  will enqueue a task into redis

* actionhero console
  will open an interactive CLI with the API object in scope.
  this is sometimes called a REPL

* actionhero link --name=[pluginName]
  will link the actions, tasks, initializers, etc from a plugin into your top-level project
  normally, you will have first installed the plugin via `npm install myPlugin`

More Help & the actionhero documentation can be found @ http://www.actionherojs.com
```

# Localization Tooling
(by @evantahler via https://github.com/evantahler/actionhero/pull/756)

This allows localization of responses to connections and the api server logs using the [i18n](https://github.com/mashpie/i18n-node) node module.

## Locale files
- When running actionhero with `api.config.i18n.updateFiles = true`, you will see actionhero generate a 'locales' folder at the top level of your project which will contain translations of all strings in your project with are passed though the new localization system.  This includes all uses of `connection.localize` and `api.log`.  
  - be sure to use sprintf-style string interpolation for variables!
- From here, it is an easy matter to change the strings, per locale, to how you would like them presented back in your application
- disable `api.config.i18n.updateFiles` if you do not want this behavior. 

## Determining connection locales
Since every actionhero implementation is unique, we cannot ship with a "guess" about how to determine a given connection's locale. Perhaps you have an HTTP server and you can trust your client's `accept-language` headers.  Or perhaps you run your API under a number of different host names and you can presume locale based on them.   Whatever the case, you need to create a synchronous method in an initializer which will be called when each connection connects to return its locale.  

For example, I may have an initializer in my project like this:
```js
module.exports = {
  initialize: function(api, next){
    api.customLocalization = {
      lookup: function(connection){
        var locale = 'en';
        if(connection.type === 'web'){
          if(connection.rawConnection.req.headers.host === 'usa.site.com'){ locale = 'en-US'; }
          if(connection.rawConnection.req.headers.host === 'uk.site.com'){  locale = 'en-GB'; }
          if(connection.rawConnection.req.headers.host === 'es.site.com'){  locale = 'es-ES'; }
          if(connection.rawConnection.req.headers.host === 'mx.site.com'){  locale = 'es-MX'; }
        }

        return locale;
      }
    }


    next();
  }
}
```

To tell the i18n to use this method with a new connection, set `api.config.i18n.determineConnectionLocale = 'api.customLocalization.lookup'`

## New Config Settings
- `config/errors.js` has been completely redone to take advantage of `connection.localize`
- `config/i18n`:

```js
exports.default = {
  i18n: function(api){
    return {
      // visit https://github.com/mashpie/i18n-node to see all configuration options
      // locale path can be configired from within ./config/api.js
      locales: ['en'],
      fallbacks: {
        'en-US': 'en',
      },

      updateFiles: true,

      // this will configure logging and error messages in the log(s)
      defaultLocale: 'en',

      // the name of the method by which to determine the connection's locale
      // by default, every request will be in the 'en' locale
      // this method will be called witin the localiazation middleware on all requests
      determineConnectionLocale: 'api.i18n.determineConnectionLocale',
    }
  }
};

exports.procution ={
  i18n: function(api){
    return {
      updateFiles: false
    }
  }
}

```

## Connection Methods:
- `connection.localize(string)` or `connection.localize([string-with-interpolation, value])`
  - Allows you to interpolate a string based on the connection's current locale.  For example, say in an action you wanted to respond with `data.response.message = connection.localize('the count was %s', 4);` In your locale files, you would define `the count was %s` in every language you cared about, and not need to modify the action itself at all.

## Logger Updates:
- Just like connections, the actionhero logger itself now has localization interpolation built in.  Note how in the new settings you set which locale you want the server's logs to be expressed in.  Using that, you can now use sprintf-style string interpolation as the first argument of `api.log()`
  - You might want to log the message `api.log(['The time is %s', new Date()], 'alert')`.  
  - Changing the translation of the string `The time is %s` in your locale files would apply to the logger as well!
  - You can of course continue to log plain strings as we have been with the logger as well.

# Linked Connections

- create the binary command `actionhero link --name=name_of_plugin`
- remove the plugin-specific code from the config initializer 
- remove the old plugin tests

## Plugin Registration

When running `actionhero link --name=name_of_plugin`, we look though your `api.config.general.paths.plugin` (which contains `./node_modules` by default) for the named plugin.  If we find it, we symlink in the actions, initializers, tasks, and servers found there.  These files will then be "present" in your project via that linkfile.  A linkfile is a simple file with the name of `{pluginName}.link` in the top level folder of that type of thing (like actions or tasks)

For example, if you install `myPlugin` via `npm install --save myPlugin` into your project's `./node_modules` folder, and then `npm run actionhero link --name myPlugin`, this will work perfectly.  

Files found in the `/config` directory of the plugin will be copied into your project's top level `config` directory.  This is under the assumption that you will want to modify these settings.  Config files within a plugin itself will not get sourced.

A previous version of this tool used OS symlinks, but that was removed to enable cross-platform compatibility. 

## Nested Plugins (TODO)

The good news is that we can count on NPM to properly (relatively) install any dependent plugins as a child in their local `node_modules`, so it shouldn't be too hard. 

## Notes for Plugin Authors
- There should be no more postinstall scripts needed
- Remember that any files in `/config` will get automatically copied into the user's project when they run `actionhero link`.  Make sure that you have some good comments!
- Be sure that your plugin is OS-independant.  Use `path.sep` for file path separators and things like that.  Use relative paths for everything in your plugin.

# Remove Domains + actionDomains from the project.

- Domains are to be removed from node.js in the near future: https://nodejs.org/api/domain.html, so they should be removed from actionhero as well.
- When Running in actionCluster, bubble up uncaught errors to cluster so they can be logged, but the underlying server instance will still crash... as it is supposed to. 
- by @evantahler via https://github.com/evantahler/actionhero/pull/736

# Remove Grunt; more Actionhero-binary commands 
- Remove all trace of Grunt from the project.  This will break any project which relies on Grunt.
- Move `actions `, `enqueueTask`, and `console` to the actionhero binary
- by @evantahler via https://github.com/evantahler/actionhero/pull/739

# Input methods have `api` set as `this`.
- Allows reference to the `api` object within input operations like `formatter` and `validator`
- `this` is set to the `api` object within these sync methods
- by @evantahler via https://github.com/evantahler/actionhero/pull/749

# Misc
- Cluster ReWrite + Graceful reboot fixes 
  - The `startCluster` module has been re-written to improve speed, stability, and legibility
  - In the process, graceful restart `SIGHUP` has been fixed to wait on each worker to fully return before moving on to restart the next one (with a timeout)
  - This also simplifies down the set of cluster options (see above):
  - By @evantahler via https://github.com/evantahler/actionhero/pull/752
- `HEAD` http verb is allowed in the router
  - by @evantahler via https://github.com/evantahler/actionhero/pull/751
- Major project formatting (JSHint) 
  - by @l0oky via https://github.com/evantahler/actionhero/pull/742
- (Bugfix) Remove `eval` via dynamic method invocation 
  - This was previously used to exec remote RPC calls.
  - By @evantahler via https://github.com/evantahler/actionhero/pull/754
- (Bugfix) Default logger config should source api.general.paths.log
  - by @evantahler via https://github.com/evantahler/actionhero/pull/745
  - by @dangerdespain via https://github.com/evantahler/actionhero/pull/748
- (Bugfix) Keyword 'package' to 'pkg' change
  - In ES6+ `object.package` is a reserved word.  Updated. how redis/resque handles it's underlying connection to be renamed to `pkg`.
  - by @synthmeat via https://github.com/evantahler/actionhero/pull/747
- Misc documentation updates by:
  - @omichowdhury
  - @evantahler
  - @l0oky


bin/actionhero
```diff
@@ -181,8 +181,8 @@ if(binary.actions[binary.mainAction]){
     binary.actions[binary.mainAction](binary, function(){});
   } else {
     binary.log('actionhero >> ' + binary.mainAction);
-    // binary.log('project_root >> ' + path.normalize(binary.paths.project_root + '/'));
-    // binary.log('actionheroRoot >> ' + path.normalize(binary.paths.actionheroRoot + '/'));
+    binary.log('project_root >> ' + path.normalize(binary.paths.project_root + '/'), 'debug');
+    binary.log('actionheroRoot >> ' + path.normalize(binary.paths.actionheroRoot + '/'), 'debug');
```
l0oky: What do you say about following a single naming convention? `actionhero_root` for example is much easier to read because `project_root` is snake case too.

evantahler: Yep, that should be `projectRoot`

PR #742, opened 2016-01-10
Major project formatting.

l0oky:
Most of the missing semicolons added. JSHint Linter reported about 720+ of these. Fixed all of them.
Not tested. Should pass though. Lets see what `travis-ci` says.


l0oky:
Great! Tests passed. Linter complains less now. There were places where semicolons really should be like `var a = 'blab'`, however, most of them were in a closed block like this `{var a = 'blab'}`.

evantahler:
WOW!
This is full-on OCD mode.  This is awesome!

PR #748, opened 2016-01-10
Normalize the logging directory config + error fix

dangerdespain:
This commit creates the logging directory according to the API config, rather than at the project root. This also means that the created directory matches the one used in line 30 for the Winston transport.

I also replaced the next(err) callback with a thrown error since api.log is not yet defined and config files don't use callbacks.

I ran across this bug while toying around with different config/api.js path configurations for different AH project directory structures.

### DIFF
--- a/config/logger.js
+++ b/config/logger.js
@@ -18,10 +18,11 @@ exports.default = {
 
     // file logger
     try{
-      fs.mkdirSync('./log');
+      fs.mkdirSync(api.config.general.paths.log[0]);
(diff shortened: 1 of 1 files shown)


evantahler:
Looks like you *just* missed this one I did :D

https://github.com/evantahler/actionhero/pull/745

evantahler:
good fix though!

dangerdespain:
Hahaha :+1:

PR #749, opened 2016-01-12
input methods have `api` set as `this`

evantahler:
Since `formatter`, `validator`, and `default` are sync methods, we can change their `this` to be API.  That will be handy.

### DIFF
--- a/initializers/actionProcessor.js
+++ b/initializers/actionProcessor.js
@@ -193,7 +193,7 @@ module.exports = {
         // default
         if(self.params[key] === undefined && props.default !== undefined){
           if(typeof props.default === 'function'){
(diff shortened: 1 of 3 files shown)


paulxtiseo:
When I use a `console.log(this)` in the input validator, I get the action, not the api, object. Is this what was intended? I have AH v13.3.1.

evantahler:
@paulxtiseo are you certain?  It's pretty clearly calling `call` with `api` as the scope https://github.com/evantahler/actionhero/blob/master/initializers/actionProcessor.js#L196-L215

paulxtiseo:
My apologies. I was using a fat arrow for the validator function.

evantahler:
... adding new meaning to `keep it lean`

PR #753, opened 2016-01-15
Change connection to data.connection in docs

omichowdhury:
I think it's clearer to be explicit in this context.


evantahler:
makes sense!  
Can you please make this change against the `v13.0.0` branch rather than master?

Thanks!

PR #761, opened 2016-01-21
fix symlink creation on windows

S3bb1:
This pull request fixes the symlink creation on windows.

See https://nodejs.org/api/fs.html#fs_fs_symlink_target_path_type_callback

fs.symlink(target, path[, type], callback)

The type argument can be set to 'dir', 'file', or 'junction' (default is 'file') and is only available on Windows (ignored on other platforms


evantahler:
awesome! Thanks!

PR #762, opened 2016-01-29
fix(initializers): paramAdd with no params raises TypeError

djkilgore:
Using paramAdd over tcp socket with no arguments throws TypeError

/node_modules/actionhero/initializers/connections.js:200
          if(words[0].indexOf('=') >= 0){
                     ^

TypeError: Cannot read property 'indexOf' of undefined


evantahler:
awesome, thanks!
I'll copy this into the upcoming v13 branch as well

PR #775, opened 2016-02-19
Package keyword issues

krishnaglick:
Older actionhero installs used redis.package instead of redis.pkg (12.2.0 at least) so this should ensure backwards compatibility.

### DIFF
--- a/initializers/redis.js
+++ b/initializers/redis.js
@@ -18,7 +18,7 @@ module.exports = {
     };
 
     if(api.config.redis.package && !api.config.redis.pkg) {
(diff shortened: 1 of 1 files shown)


evantahler:
I like this, but can you add a warning (`api.log`) which will show a deprecation warning?  Also, this will work for api.redis, but not for the task system which re-sources the same config.   To be complete, you'll need to make a similar modification to the https://github.com/taskrabbit/node-resque project

krishnaglick:
Roger, I'll throw up the necessary changes tomorrow. Thanks for letting me know!

evantahler:
fyi @synthmeat

evantahler:
not dumb at all!

krishnaglick:
I had missed a ;
Felt dumb :P

Issue #772, opened 2016-02-18
Using the `ActionheroClient.prototype.actionWeb` method


running-coder:
Hey, my case is pretty simple, at some point in my application the user is not connected with the WS so the condition falls back inside the `actionWeb` method. I was playing around trying to understand how change the request type to `GET` and found out the `httpMethod` option.

[params.httpMethod](https://github.com/evantahler/actionhero/blob/v13.0.2/client/actionheroClient.js#L194)

The `xmlhttp.send(JSON.stringify(params));` works for a `POST` request but does not work with a `GET` unless I'm missing something.

My code looks like:
``` javascript
client.action("gameHistoryViewAll", {httpMethod: 'GET', gameId: 42}, function (apiResponse) {
    console.log(apiResponse)
});
```

![2016-02-18_1341](https://cloud.githubusercontent.com/assets/7377382/13154133/7d076220-d645-11e5-9e35-82aefeb6695b.png)

![2016-02-18_1342](https://cloud.githubusercontent.com/assets/7377382/13154137/8084fbec-d645-11e5-9cc4-dbdbec89c8de.png)

As you can see it's missing the `gameId` parameter inside the query string

evantahler:
If you change the method to `GET`, you actually need to serialize the params to URL variables, IE: `site.com?thing=stuff`.  

Send in a PR which checks for GET and modifies the URL (which is what would be needed)

running-coder:
I've slightly modified the method to include the `GET` case. The `toUpperCase` is not mandatory but I have the habit of always doing strict comparisons

``` javascript
ActionheroClient.prototype.actionWeb = function(params, callback) {
  var xmlhttp = new XMLHttpRequest();
  xmlhttp.onreadystatechange = function () {
    var response;
    if(xmlhttp.readyState === 4) {
      if(xmlhttp.status === 200) {
        response = JSON.parse(xmlhttp.responseText);
      }else{
        try{
          response = JSON.parse(xmlhttp.responseText);
        }catch(e){
          response = { error: {statusText: xmlhttp.statusText, responseText: xmlhttp.responseText} };
        }
      }
      callback(response);
    }
  };

  var method = (params.httpMethod || 'POST').toUpperCase();
  var url = this.options.url + this.options.apiPath + '?action=' + params.action;

  if (method === 'GET') {
    for (var param in params) {
      if (~['action', 'httpMethod'].indexOf(param)) continue;
      url += '&' + param + '=' + params[param];
    }
  }

[...]

PR #776, opened 2016-02-19
#772 - Using the `ActionheroClient.prototype.actionWeb` with GET method

running-coder:
Referring to #772, will enable to have the `GET` query params inside the url in case of a client action is called using the actionWeb method

ex:
``` javascript
client.action("myActionName", {httpMethod: 'GET', param1: value1, param2: value2}, function (apiResponse) {
    console.log(apiResponse)
});

// will result to:

{webApiUrl}/api?action=myActionName&param1=value1&param2=value2
```

### DIFF
--- a/client/actionheroClient.js
+++ b/client/actionheroClient.js
@@ -191,8 +191,16 @@ ActionheroClient.prototype.actionWeb = function(params, callback) {
     }
   };
 
-  var method = params.httpMethod || 'POST';
+  var method = (params.httpMethod || 'POST').toUpperCase();
   var url = this.options.url + this.options.apiPath + '?action=' + params.action;
+
+  if (method === 'GET') {
+    for (var param in params) {
+      if (~['action', 'httpMethod'].indexOf(param)) continue;
+      url += '&' + param + '=' + params[param];
+    }
+  }
+
   xmlhttp.open(method, url, true);
   xmlhttp.setRequestHeader('Content-Type', 'application/json');
   xmlhttp.send(JSON.stringify(params));


evantahler:
Looks like the right implementation!
Please add a test (probably in /test/servers/web) so we can lock this in and ensure it won't get changed away in the future.

running-coder:
Sure, I had some difficulties running the test under windows and found out it have to be written in a special way

``` javascript
"test": "SET NODE_ENV=test & mocha",
```

I'll commit a test as soon as I get some time to learn / work on it

running-coder:
I'm trying to reach the `ActionheroClient().action` function but having difficulties... (imo it seems pointless to call the `request` module since the issue to be tested is how the request gets formed). I saw `api.specHelper.runAction` method in other test files but for some reason api.specHelper is undefined on my setup (lets blame windows). Maybe this task gets beyond of my testing framework understandings...

``` javascript
it('will create a GET request with multiple parameters', function(done){
    api.specHelper.runAction('x', {httpMethod: 'GET', thing: 'stuff', moreThings: 'moreStuff'}, function(response){
      // Not sure how to validate the request method
      response.requesterInformation.action.should.equal('x');
      response.requesterInformation.thing.should.equal('stuff');
      response.requesterInformation.moreThings.should.equal('moreStuff');
      done();
    });
  });
```

evantahler:
The [spec helper](http://www.actionherojs.com/docs/#test-methods) is actually a request-agnostic way to test actions (http, ws, etc) and won't help here because we want to test the actual mechanics of the HTTP client.  

I'll merge this in now and add a test.  Thanks!

running-coder:
Cool! Let me know how the test looks like (I'm very curious with the result)
Thanks

evantahler:
I was able to source the client lib much like how the websocket tests do: https://github.com/evantahler/actionhero/blob/master/test/servers/websocket.js#L14-L32

However, I realized that `XMLHttpRequest` isn't a thing.  I found a good library that has a node proxy for it (https://github.com/driverdan/node-XMLHttpRequest) but that seemed kind of crazy to require... so guess we aren't testing this one :D

running-coder:
haha! alright

PR #777, opened 2016-02-19
wrapping a 'package', get it?

synthmeat:
Same as for node-resque.


evantahler:
hilarious.  And thanks!

PR #781, opened 2016-02-24
NODE_ENV fix for Windows

krishnaglick:
This gets tests running on Windows but I don't know if it breaks Unix systems.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,8 +61,8 @@
   "scripts": {
     "postinstall": "echo 'To generate a new actionhero project, run \"node ./node_modules/.bin/actionhero generate\"'",
     "help": "node ./bin/actionhero help",
-    "test": "NODE_ENV=test mocha",
-    "test-travis": "NODE_ENV=test ./node_modules/istanbul/lib/cli.js cover ./node_modules/.bin/_mocha",
+    "test": "SET NODE_ENV=test & mocha",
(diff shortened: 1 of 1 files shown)


krishnaglick:
Well, it looks like it does. :/

evantahler:
It looks like folks have gone through a ton of hackery to get stuff like this working, as evidenced by packages like https://www.npmjs.com/package/cross-env and this conversation about NPM itself https://github.com/npm/npm/issues/10575

I guess the right thing to do here is to make a single node script in the `/test/run-test.js` directory.  It would be a little JS file which will detect the OS (check `process.platform`), and then uses the a child process to actually call out to mocha.   This is vaguely what NPM recommends in this note https://github.com/npm/npm/issues/2800

evantahler:
This is also why the fancy packages use makefiles for this too https://github.com/mochajs/mocha/blob/master/Makefile

I was going to suggest a bash script to wrap this all... but then again... windows :(

krishnaglick:
I'm gonna say this isn't worth the work unless some adventurous young dev wants to go through the trouble.
Maybe mark it as a low priority error?

evantahler:
Marked!

evantahler:
closing in favor of https://github.com/evantahler/actionhero/pull/790

Issue #780, opened 2016-02-24
Live reloading of changed Config fails


S3bb1:
When hot reloading changed Configs, ActionHero crashes with error message : 

```
2016-02-23T18:32:43.553Z - info:  0=*** rebooting due to config change (%s) ***, 1=/project/api/config/sequelize.js
/project/api/node_modules/actionhero/initializers/config.js:96
      api.commands.restart.call(api._self);
                  ^

TypeError: Cannot read property 'restart' of undefined
    at rebootCallback (/project/api/node_modules/actionhero/initializers/config.js:96:19)
    at /project/api/node_modules/actionhero/initializers/config.js:27:15
    at nextTickCallbackWith0Args (node.js:452:9)
    at process._tickDomainCallback (node.js:422:13) 
```

This issue comes from the line https://github.com/evantahler/actionhero/blob/master/initializers/config.js#L19

It seems that watchFile does strange things with the current scope. 

When the watchFile is replaced with fs.watch, all works fine.
```
fs.watchFile(file, function(event, filename){
          if(event === 'change' && api.config.general.developmentMode === true){
```

Found a note in the Node.JS wiki:

> Note: fs.watch() is more efficient than fs.watchFile and fs.unwatchFile. fs.watch should be used instead of fs.watchFile and fs.unwatchFile when possible.

evantahler:
I can't reproduce this at all.  I can tell that you are using an older version of node here because the string interpolation isn't working.  Can you try again on node 4+ with the latest version of actionhero?

```
2016-02-24T18:18:28.562Z - info: actionhero >> start
2016-02-24T18:18:28.656Z - notice: *** starting actionhero ***
2016-02-24T18:18:28.713Z - warning: running with fakeredis
2016-02-24T18:18:28.724Z - info: actionhero member 192.168.80.162 has joined the cluster
2016-02-24T18:18:28.846Z - notice: pid: 90064
2016-02-24T18:18:28.847Z - notice: server ID: 192.168.80.162
2016-02-24T18:18:28.857Z - notice: Starting server: web
2016-02-24T18:18:28.865Z - notice: Starting server: websocket
2016-02-24T18:18:29.460Z - notice: environment: development
2016-02-24T18:18:29.461Z - notice: *** Server Started ***

2016-02-24T18:18:31.855Z - info: *** Rebooting due to initializer change (/Users/evantahler/Dropbox/Desktop/actionhero_tmp/initializers/test.js) ***

[...]

PR #783, opened 2016-02-25
only reload server (in development mode) when running

evantahler:
Only enable the callbacks to `api.watchFileAndAct` when `api.running === true`.  This helps us in 2 ways:
- prevents badly formatted or error'd initializers from constantly getting reloaded until we get at least one clean boot of the server. 
- solves #780, which was caused by both the actionhero server and the actionhero binary (which wraps the actionhero server) *both* watching the file... and trying to react to the change.

### DIFF
--- a/initializers/config.js
+++ b/initializers/config.js
@@ -7,22 +7,42 @@ module.exports = {
   loadPriority:  0,
   initialize: function(api, next){
 
+    // api.env
+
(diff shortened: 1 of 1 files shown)


evantahler:
solves https://github.com/evantahler/actionhero/issues/780

PR #786, opened 2016-02-26
Add rawConnection.serializer

LLFourn:
refactors out JSON.stringify into the default value of
rawConnection.serializer. This allows it to be customized per-request.

PS After looking around the code base a bit more I think actionhero is probably too tied to JSON in general for my (rather odd) needs. I probably won't be using it (apart from stealing ideas). But here's the PR anyway, it might be useful if you want to customize the JSON output for socket or something.

Cheers

### DIFF
--- a/servers/socket.js
+++ b/servers/socket.js
@@ -79,7 +79,7 @@ var initialize = function(api, options, next){
       }
     }
     try {
-      connection.rawConnection.write(JSON.stringify(message) + '\r\n');
+      connection.rawConnection.write(connection.rawConnection.serializer(message) + '\r\n');
     } catch(e){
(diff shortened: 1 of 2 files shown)


evantahler:
Thanks for the PR!

I'm not going to merge this for a few reasons:
- Rather than just adding a method, this should be user configurable.  Do you expect users to just meta-program over the existing methods?
- Should the serializer be set per connection of such a thing can be used for more than one server type?  If so, this belongs in the connection, not the server (and the sever would then invoke `connection.serialize` or something)
- no docs
- no tests for added features.

That said, thanks for pointing out that we can do this!

evantahler:
Shame that actionhero can't help you out... good luck!

PR #789, opened 2016-03-02
Support /public filenames with encoded spaces in URL

bluesunrise:
Support /public filenames with encoded spaces in URL. The URL must be properly encoded, for example, note the %20 and +:

http://localhost:8080/sky%20with+space.jpg

### DIFF
--- a/servers/web.js
+++ b/servers/web.js
@@ -447,10 +447,10 @@ var initialize = function(api, options, next){
         connection.params.file = connection.params.file + api.config.general.directoryFileType;
       }
       try {
-        connection.params.file = decodeURIComponent(connection.params.file).replace(/\++/g, ' ');
+        connection.params.file = decodeURIComponent(connection.params.file);
       }
       catch(e) {
-        connection.error = new Error('There was an error decode URI.');
+        connection.error = new Error('There was an error decoding URI: ' + e);
       }
       callback(requestMode);
     }
--- a/servers/web.js
+++ b/servers/web.js
@@ -446,6 +446,12 @@ var initialize = function(api, options, next){
       if(connection.params.file === '' || connection.params.file[connection.params.file.length - 1] === '/'){
         connection.params.file = connection.params.file + api.config.general.directoryFileType;
       }
+      try {
+        connection.params.file = decodeURIComponent(connection.params.file).replace(/\++/g, ' ');
+      }
+      catch(e) {
+        connection.error = new Error('There was an error decode URI.');
+      }
       callback(requestMode);
     }
 
--- a/test/servers/web.js
+++ b/test/servers/web.js
@@ -909,6 +909,45 @@ describe('Server: Web', function(){
 
     });
 
+    describe('spaces in URL with public files', function() {
+
+      var source = __dirname + '/../../public/logo/sky.jpg'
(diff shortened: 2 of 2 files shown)


test/servers/web.js
```diff
@@ -909,6 +909,45 @@ describe('Server: Web', function(){
...
+        request.get(url + '/sky%20with+space.jpg', function (err, response) {
+          response.statusCode.should.equal(200)
+          response.body.should.be.an.instanceOf(Object);
+          response.headers['content-type'].should.equal('image/jpeg');
+          done();
+        });
+      });
+
+      it('will capture bad encoding in URL and return NOT FOUND', function (done) {
+        request.get(url + '/sky%20%with+space.jpg', function (err, response) {
```
submitteddenied: I'd prefer that this test was more obviously broken, the extra `%` gets lost pretty easily

servers/web.js
```diff
@@ -446,6 +446,12 @@ var initialize = function(api, options, next){
       if(connection.params.file === '' || connection.params.file[connection.params.file.length - 1] === '/'){
         connection.params.file = connection.params.file + api.config.general.directoryFileType;
       }
+      try {
+        connection.params.file = decodeURIComponent(connection.params.file).replace(/\++/g, ' ');
```
submitteddenied: I'm happy to be overruled here, but I don't like the addition of handling `+`.

If `decodeURIComponent` doesn't convert `+` to space, it's better to fix offending clients than add extra handling here, IMO.

test/servers/web.js
evantahler: :+1:

servers/web.js
evantahler: :+1: to that.  `decodeURIComponent ` will convert %20 to space... and isn't that all we need?
```js
> decodeURIComponent('word%20other%20word')
'word other word'
```

bluesunrise: The + support was added because I've just usually done that in the past. There are lots of examples of URLs with + in them for spaces. My case doesn't need the pluses, feel free to remove it

test/servers/web.js
bluesunrise: Would that be more obvious?:

/sky%20%%%%%%%with+space.jpg'

servers/web.js
evantahler: You've got to update your PR :D

bluesunrise:
> You've got to update your PR :D

updated it with removal of + conversion to space, now it only converts %20
also made the failed test case more obvious with lots of %%%%%%

hutchike:
I'd +1 using decodeURIComponent

evantahler:
Awesome!

evantahler:
I'm going to update the test not to use `/tmp` (windows users hate that), but otherwise, great PR!

bluesunrise:
> I'm going to update the test not to use /tmp (windows users hate that), but otherwise, great PR!

I was about to ask you about /tmp on Windows. 
There is another test in test/servers/web.js that also uses /tmp, see line 577, 585

evantahler:
@bluesunrise we'll do some work on it here https://github.com/evantahler/actionhero/pull/790

PR #794, opened 2016-03-16
fix cluster workers IDs > 10

evantahler:
silly javascript... defaulting to string comparisons...


evantahler:
solves https://github.com/evantahler/actionhero/issues/791

PR #795, opened 2016-03-19
Async test for the status action was always passing since the done param was missing

running-coder:
Simply correcting the async test example for the status action


evantahler:
good find! Thanks!

PR #799, opened 2016-03-24
Added "use strict" to templates

ndcollins:
Nothing major, but saves a hassle of a step if I want to use ES6 or ES7 features in Actionhero.

### DIFF
--- a/bin/templates/action.js
+++ b/bin/templates/action.js
@@ -1,3 +1,4 @@
+"use strict";
 exports.action = {
   name:                   '%%name%%',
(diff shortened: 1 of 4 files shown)


evantahler:
For those who want to lear more about `"use strict"` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode

evantahler:
Thanks!

evantahler:
You've inspired me! We might as well do this in the core too https://github.com/evantahler/actionhero/pull/800

PR #804, opened 2016-04-01
Templating Change

krishnaglick:
1. Newline at beginning of file. Stylistic more than anything.
2. Space before curly brace on functions.
3. let instead of var

Thoughts?

### DIFF
--- a/bin/templates/action.js
+++ b/bin/templates/action.js
@@ -1,3 +1,4 @@
+
 'use strict';
 
 exports.action = {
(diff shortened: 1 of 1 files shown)


bin/templates/action.js
```diff
@@ -12,8 +13,8 @@ exports.action = {
 
   inputs: {},
 
-  run: function(api, data, next){
-    var error = null;
+  run: function(api, data, next) {
+    let error = null;
```
evantahler: Can you confirm that node v4 is comfortable with the `let` syntax?  I thought that came with node v5...

krishnaglick:
Using 4.4.2 I ran:

```
'use strict';
let path = require('path');
let glob = require('globby');

glob(path.resolve('node_modules/actionhero/bin/templates/**'))
.then(console.log);

```
And it ran fine. This is a test.js file I have in my app.

evantahler:
ok, cool!

PR #806, opened 2016-04-01
eslint npm script, instructions, and a single style fix

synthmeat:
- now `eslint` dependency makes sense
- it's easy to check style - just run `npm run lint`
- added .eslintignore to not check troublesome scripts (client, templates, site, ...)

A side note - there really was no need for Grunt. Good call removing it.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,5 @@
+bin/templates/
+client/
+public/
+site/
+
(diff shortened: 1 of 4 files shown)


evantahler:
nice.

evantahler:
so... we have a lot of problems... can you exclude node-modeules?

synthmeat:
Hm, `node_module` is [ignored by default](http://eslint.org/docs/user-guide/configuring#ignoring-files-and-directories), and `npm run lint` gives me no errors.

evantahler:
wasn't for me... https://github.com/evantahler/actionhero/commit/56d61a07143c77098578c988c7bc6a43840527df Just in case

evantahler:
it's all good now!

PR #808, opened 2016-04-04
maintain plugin config-file sub folders when linking 

evantahler:


evantahler:
solves https://github.com/evantahler/actionhero/issues/807

PR #817, opened 2016-04-11
localized goodbyeMessage for socket clients

evantahler:


evantahler:
helps solve https://github.com/evantahler/actionhero/issues/816

PR #821, opened 2016-04-14
redis HA docs

evantahler:


evantahler:
This fixes up https://github.com/evantahler/actionhero/pull/819

evantahler:
And here we are! http://www.actionherojs.com/docs/#redis-high-availability-configurations

PR #823, opened 2016-04-24
Added dynamic delimiter and maxSocketDataLength feature.

l0oky:
I also added some tests. _Maybe_ needs some more testing.

### DIFF
--- a/config/servers/socket.js
+++ b/config/servers/socket.js
@@ -5,15 +5,19 @@ exports['default'] = {
         enabled: false,
         // TCP or TLS?
         secure: false,
-        // passed to tls.createServer if secure=true. Should contain SSL certificates
+        // Passed to tls.createServer if secure=true. Should contain SSL certificates
         serverOptions: {},
         // Port or Socket
         port: 5000,
-        // which IP to listen on (use 0.0.0.0 for all)
+        // Which IP to listen on (use 0.0.0.0 for all)
         bindIP: '0.0.0.0',
-        // Enabple TCP KeepAlive pings on each connection?
+        // Enable TCP KeepAlive pings on each connection?
         setKeepAlive: false,
-        // what message to send down to a client who requests a `quit`
+        // Delimiter string for incoming messages
+        defaultDelimiter: '\n',
+        // Maximum incoming message string length in Bytes (use 0 for Infinite)
+        maxSocketDataLength: 0,
+        // What message to send down to a client who requests a `quit`
         goodbyeMessage: 'Bye!'
       };
     }
--- a/servers/socket.js
+++ b/servers/socket.js
@@ -48,6 +48,10 @@ var initialize = function(api, options, next){
       });
     }
 
+    if(typeof api.config.servers.socket.defaultDelimiter !== 'string'){
+      return next(new Error('Cannot start socket server @ Wrong delimiter type, delimiter must be string.'));
+    }
+
     server.server.on('error', function(e){
       return next(new Error('Cannot start socket server @ ' + options.bindIP + ':' + options.port + ' => ' + e.message));
     });
@@ -107,6 +111,11 @@ var initialize = function(api, options, next){
     connection.params = {};
 
     var parseLine = function(line){
+      var blen = byteLength(line);
+      if(blen > api.config.servers.socket.maxSocketDataLength && api.config.servers.socket.maxSocketDataLength !== 0){
+        server.log('socket error: incoming data length is too big ('+api.config.servers.socket.maxSocketDataLength+'<'+blen+')', 'error');
+        return;
+      }
       if(line.length > 0){
         // increment at the start of the request so that responses can be caught in order on the client
         // this is not handled by the genericServer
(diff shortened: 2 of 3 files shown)


config/servers/socket.js
```diff
@@ -5,15 +5,19 @@ exports['default'] = {
...
+        // Which IP to listen on (use 0.0.0.0 for all)
         bindIP: '0.0.0.0',
-        // Enabple TCP KeepAlive pings on each connection?
+        // Enable TCP KeepAlive pings on each connection?
         setKeepAlive: false,
-        // what message to send down to a client who requests a `quit`
+        // Delimiter string for incoming messages
+        defaultDelimiter: '\n',
+        // Maximum incoming message string length in Bytes (use 0 for Infinite)
+        maxSocketDataLength: 0,
```
l0oky: This feature still doesn't care much about that potential memory leak I talked about, though...

test/servers/socket.js
```diff
@@ -226,6 +229,54 @@ describe('Server: Socket', function(){
...
+      api.config.servers.socket.defaultDelimiter = '^]';
+      makeSocketRequest(client, JSON.stringify({action: 'status'}), function(response){
+        response.context.should.equal('response');
+        done();
+      }, '^]');
+    });
+
+    // Errors at socket server start:
+    //  Cannot start socket server @ Wrong delimiter type, delimiter must be string.
+    it('will error If data delimiter is not string');
```
l0oky: I didn't know how to test this. This is straight forward, I guess It works.

config/servers/socket.js
```diff
@@ -5,15 +5,19 @@ exports['default'] = {
...
         port: 5000,
-        // which IP to listen on (use 0.0.0.0 for all)
+        // Which IP to listen on (use 0.0.0.0 for all)
         bindIP: '0.0.0.0',
-        // Enabple TCP KeepAlive pings on each connection?
+        // Enable TCP KeepAlive pings on each connection?
         setKeepAlive: false,
-        // what message to send down to a client who requests a `quit`
+        // Delimiter string for incoming messages
+        defaultDelimiter: '\n',
```
evantahler: I think this should just be `delimiter`, as there is no default... it's just the value.

servers/socket.js
```diff
@@ -107,6 +111,11 @@ var initialize = function(api, options, next){
     connection.params = {};
 
     var parseLine = function(line){
+      var blen = byteLength(line);
+      if(blen > api.config.servers.socket.maxSocketDataLength && api.config.servers.socket.maxSocketDataLength !== 0){
```
evantahler: to make this faster, don't get the `byteLength(line);` until after you've already checked that `api.config.servers.socket.maxSocketDataLength`.

servers/socket.js
```diff
@@ -107,6 +111,11 @@ var initialize = function(api, options, next){
...
+        server.log('socket error: incoming data length is too big ('+api.config.servers.socket.maxSocketDataLength+'<'+blen+')', 'error');
```
evantahler: I don't think this is the correct behavior.  Shouldn't the client get an error message?

servers/socket.js
```diff
@@ -48,6 +48,10 @@ var initialize = function(api, options, next){
       });
     }
 
+    if(typeof api.config.servers.socket.defaultDelimiter !== 'string'){
```
evantahler: No need for this if you `String(api.config.servers.socket.defaultDelimiter)` as noted below

servers/socket.js
```diff
@@ -119,13 +128,17 @@ var initialize = function(api, options, next){
       if(checkBreakChars(chunk)){
         connection.destroy();
       }else{
+        // Replace all carriage returns with newlines.
         connection.rawConnection.socketDataString += chunk.toString('utf-8').replace(/\r/g, '\n');
         var index;
-        while((index = connection.rawConnection.socketDataString.indexOf('\n')) > -1){
+        var d = api.config.servers.socket.defaultDelimiter;
```
evantahler: `var d = String(api.config.servers.socket.defaultDelimiter);` for type saftey

servers/socket.js
```diff
@@ -119,13 +128,17 @@ var initialize = function(api, options, next){
...
+
+        while((index = connection.rawConnection.socketDataString.indexOf(d)) > -1){
           var data = connection.rawConnection.socketDataString.slice(0, index);
           connection.rawConnection.socketDataString = connection.rawConnection.socketDataString.slice(index + 2);
```
evantahler: The "2" here is the length of d

servers/socket.js
```diff
@@ -218,6 +231,18 @@ var initialize = function(api, options, next){
     return found;
   };
 
+  var byteLength = function(str) {
```
evantahler: I think this will be a lot faster:

```
return Buffer.byteLength(string, 'utf8')
```

test/servers/socket.js
evantahler: I wouldn't worry about this per the comments above

test/servers/socket.js
```diff
@@ -226,6 +229,54 @@ describe('Server: Socket', function(){
...
+
+    var msg = {
+      action: 'cacheTest',
+      params: {
+        key: uuid.v4(),
+        value: uuid.v4() + uuid.v4() + uuid.v4() + uuid.v4() + uuid.v4() + uuid.v4() + uuid.v4() + uuid.v4() + uuid.v4() + uuid.v4()
+      }
+    };
+    makeSocketRequest(client, JSON.stringify(msg), function(response){
+      should(response).not.be.ok();
```
evantahler: What is the response here?  Should the error message for this case be set in `config/errors.js`?

evantahler:
Good start @l0oky!  I've added some comments

servers/socket.js
l0oky: Yeah, by 2 miliseconds. I did a benchmark on a 10000 Byte string (15 ms).

servers/socket.js
l0oky: I could replace that line 119 `if(line.length > 0){` with `if(blen > 0){`. I guess that would be faster? What do you think?

l0oky: `blen` could be used intead of `line.length` at line 119? I guess we get the same effect?

l0oky:
I recently deleted my `actionhero` fork so I have to create a new pull request and close this one.

PR #826, opened 2016-04-28
Map static folders in routes.js

S3bb1:
Long time ago, i've created this issue https://github.com/evantahler/actionhero/issues/451

Now i want to provide the functionality to map static folders in the routes.js file.

You can now map a folder like this:

```javascript
get: [
  { path: '/my/special/folder', dir: __dirname + '/../my/special/folder', matchTrailingPathParts: true }
],
```

and access all files under the given "dir" path within the mapped path.

Greetings


evantahler:
First, this is a GREAT pull request!  Tests, docs... it's all here!  Well done!

Before this PR, we never allowed static files to be served from outside of the `/public` directory, as a good way to ensure safety.  Be sure that I can't request a route like `/my/special/folder/../../../config/db.js` and be shown the contents of the project's source.

S3bb1:
There is a test Case for this Problem available in the pull request

test/servers/web.js
```diff
@@ -571,6 +571,47 @@ describe('Server: Web', function(){
...
+
+      it('returns 404 for files not available in route mapped paths', function(done){
+        request.get(url + '/my/custom/route/fileNotFound.html', function(err, response){
+          response.statusCode.should.equal(404);
+          done();
+        });
+      });
+
+      it('I should not see files outside of the mapped dir', function(done){
+        request.get(url + '/my/secret/route/../config/servers/web.js', function(err, response){
```
evantahler: Wouldn't a more valid test be `/my/secret/route/../../../config/servers/web.js`?  That file really doesn't exist

S3bb1:
The Route /my/secret/route is mapped to the ah public folder. 

__dirname + '/../../public'. 

Maybe a second Test where the Access to the public folder is successful resolves the wierdness or I change the Route Name to /route/to/ah/public/folder

initializers/routes.js
```diff
@@ -32,7 +34,13 @@ module.exports = {
               }
             }
             connection.matchedRoute = route;
-            connection.params.action = route.action;
+
+            if(route.dir){
+              var requestedFile = connection.rawConnection.parsedURL.pathname.substring(route.path.length, connection.rawConnection.parsedURL.pathname.length);
+              connection.params.routesFile = path.normalize(route.dir + '/'  + requestedFile);
```
evantahler: What was the benefit of creating a new object on the connection rather than setting `connection.params.file`?

evantahler: If you do that, many of the other modifications aren't needed

evantahler: This would however require the file served to reside within `/public`

S3bb1: With this change it dont breaks the logic for the public folders

evantahler: I don't understand how that would change the public folder logic?  I *want* to enforce that we can only serve files from the public folder, so the job of `dir` within the routes is simply to help build the proper file path

S3bb1: Ah okay now I understand your wanted behavior. Okay i will change the Logic to folders mappable ONLY in their specific public folder :-)

evantahler:
Cool! Let me know when you update this!

S3bb1:
HI @evantahler 

i've changed the logic that only folders within public folders can be mapped and accessed.

The logic to check if the requested file is within a public folder in the "staticFile.js" remains as is. So no other file can be accessed.

Kind regards

evantahler:
Awesome!

PR #830, opened 2016-05-06
[Demo] Rather unsafe demonstration of hotpatching and task data patching.

gcoonrod:
# **DEMONSTRATION ONLY**

While working with production deployments of ActionHero we have run into a potentially significant issue. Namely, when upgrading our Tasks it is possible that a new version of a Task is unable to handle data coming off the queue that was originally intended for the older Task version. As a technical demonstration we have modified the Tasks initializer to write a hash (sha1) of the existing Task's `run()` function into the `params` object prior to writing to the queue. Then, when a worker pulls the job from the queue it will check for the presence of the hash, calculate the hash for the current `run()` function, and compare them. If the hashes do not match a `migrate()` function is invoked. The `migrate()` function is responsible for converting the parameters from the previously expected format into the currently expected format.

## Known Limitations

- Only one upgrade per Task can be ongoing at a time.
- The Task developer is responsible for ensuring the migration is correct
- Tasks running old code **must** be prevented from executing data in the new format

## What is `api.eval`?
As a technical exercise (in poor decision making) we've added the ability to directly modify the ActionHero code base through some Actions and creative use of `fs.writeFile` and `process.send`. YMMV

### Acknowledgements
Mastermind: @omichowdhury 
Code Monkey 1: @gcoonrod
Code Monkey 2: @glambert22
Code Monkey 3: @aibarra988


evantahler:
Nice.

PR #831, opened 2016-05-11
Slack (no more gitter)

evantahler:


evantahler:
That is correct.
We are no longer using gitter.

It's slack time.

PR #836, opened 2016-05-11
Localization framework has been cleaned up and more...

l0oky:
Now you can access your i18n object with `api.i18n`. Also removed some bloated code.


l0oky:
Not sure why build #2550.4 failed :/

Maybe that occurs from time to time?

evantahler:
Can't merge now... my bad... can you rebase @l0oky ?

evantahler:
but I like this change!

PR #837, opened 2016-05-12
I18n rewrite

l0oky:
No more ugly `api.i18n.i18n...` calls. Added `api.i18n.localize`. 
Also localized some strings in `initializers/cache.js` and `config/errors.js`.

Localizing strings is now really easy. :) I guess documentation has to be changed now.

`api.i18n.localize(message, options)`, first argument is sprintf array it can also accept a string, second argument is optional ( default is api.i18n )

### DIFF
--- a/initializers/i18n.js
+++ b/initializers/i18n.js
@@ -2,7 +2,6 @@
 
 var i18n = require('i18n');
 var path = require('path');
-var extend = require('util')._extend;
 
 module.exports = {
   loadPriority:  1,
@@ -12,7 +11,7 @@ module.exports = {
     i18n.configure(options);
     i18n.setLocale(api.config.i18n.defaultLocale);
 
-    api.i18n = extend({
+    api.i18n = Object.assign({
       // simplistic determination of locale for connection
       determineConnectionLocale: function(connection){
         // perhpas you want to look at the `accept-language` headers from a web requests
--- a/initializers/i18n.js
+++ b/initializers/i18n.js
@@ -19,7 +19,6 @@ module.exports = {
         // perhaps your API can use a certain cookie or URL to determine locale
         return api.config.i18n.defaultLocale;
       },
-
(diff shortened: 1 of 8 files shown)


l0oky:
Can we increase the timeout on some of these tests? Travis gets slow sometimes..

evantahler:
you certainly can! just do `this.timeout()` in the test, or change it globally in `test/mocha.opts`

l0oky:
This should be it. I'll be waiting for your review :)

initializers/i18n.js
```diff
@@ -2,36 +2,37 @@
...
-    api.i18n = {
-      // i18n: i18n,
+    var options = api.config.i18n;
+    options.directory = path.normalize(api.config.general.paths.locale[0]);
+    i18n.configure(options);
+    i18n.setLocale(api.config.i18n.defaultLocale);
 
+    api.i18n = extend({
       // simplistic determination of locale for connection
       determineConnectionLocale: function(connection){
```
evantahler: How would application developers override this method now?  The docs say to override this method... what is the changed behavior?  http://www.actionherojs.com/docs/#determining-connection-locales

l0oky: I guess `Object.assign({}, obj);` is more appropriate as `util._extend()` is not meant to be used outside of `nodejs` internals. 

Hmm, looks like `actionhero` documentation shouldn't change anything. `api.config.i18n` stays the same as before, `connection.localize` stays the same. What I meant by "I guess documentation has to be changed now." is probably adding the function `api.i18n.localize` to the docs.

evantahler: Oh! I didn't catch that you were making a 'copy' of `i18n = require('i18n');`.  Yeah, use `Object.assign`.

l0oky:
Did I explain the `api.i18n.localize` well? :)

evantahler:
lovely

PR #846, opened 2016-05-23
Fixes direct access of servers via api.servers.servers

TobiTonner:
Found that `api.servers.servers` was empty when returning data to the client from an action. Looks like it was defined as an Array but then subsequently is used as an Object.
Changing it to Object allows us to stream results directly from an action to the client, e.g. `api.servers.servers.web.sendFile(connection, error, fileStream, mime, length, lastModified)`.

### DIFF
--- a/initializers/servers.js
+++ b/initializers/servers.js
@@ -10,7 +10,7 @@ module.exports = {
   initialize: function(api, next){
 
     api.servers = {};
-    api.servers.servers = [];
+    api.servers.servers = {};
 
     // Load the servers
 


evantahler:
Thank you for the fix!  

Due to some of the crazy quirks in javascript, you can actually access "named" properties of an array, ie: 

```js
node
> a = []
[]
> a.thing = 'stuff'
'stuff'
> a
[ thing: 'stuff' ]
> a.thing
'stuff'
> Array.isArray(a)
true
> typeof a
'object'
> a.length
0
```

But yes, `api.servers.servers` should be an object.

As a note, you can call `data.connection.sendFile()` from within your actions... you shouldn't need to use the `server`.

evantahler:
For reference: http://www.actionherojs.com/docs/#sending-files-from-actions

TobiTonner:
Thanks for merging.
Yes, I'm already using the `data.connection.sendFile()` when I'm returning actual files from the file system.
In this case, however, I wanted to stream results from a database query (thousands of rows) back to the client. Using `api.servers.servers.web.sendFile(connection, error, fileStream, mime, length, lastModified)` seems to work well - or are there any other options to pipe data back to the client?

evantahler:
Ah! That's a pretty cool use-case!  Would you mind sharing more about what you are working on (via Blog Post perhaps?)

You can also always use `data.connection.rawConnection.res`, but that's what `api.servers.servers.web.sendFile` handles under the hood for you anyway... I would say keep using it.

PR #848, opened 2016-05-23
fix crash due to missing logger in linked config setup

evantahler:
solves https://github.com/evantahler/actionhero/issues/847


gcoonrod:
👍

PR #853, opened 2016-05-25
[RFC] Task Middleware Demo

gcoonrod:
This PR demonstrates a very basic implementation of middleware for tasks. Its very nearly a direct code copy from the action middleware implementation. 

I'm looking for comments and suggestions before I go any further.

### DIFF
--- a/initializers/demoMiddleware.js
+++ b/initializers/demoMiddleware.js
@@ -1,0 +1,25 @@
+'use strict';
+var _ = require('lodash');
+
+module.exports = {
+  loadPriority:  1000,
+  initialize: function(api, next){
+    api.demoMiddleware = {
+      name: "demoMiddleware",
+      priority: 999,
+      preProcessor: function(data, next){
+        api.log("In Preprocessor!", "info");
+        next(null, true);
+      },
+      postProcessor: function(data, next){
+        api.log("In Postprocessor!", "info");
+        api.log("Data keys: ", "info", _.keys(data));
+        api.log("Worker result", "info", data.worker.result);
+        next(null, true);
+      }
+    };
+    api.tasks.addMiddleware(api.demoMiddleware);
+
+    next();
+  }
+};
--- a/initializers/taskMiddleware.js
+++ b/initializers/taskMiddleware.js
@@ -1,0 +1,71 @@
+'use strict';
+var async = require('async');
+var Util = require('util');
+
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -37,6 +37,7 @@
     "i18n": "^0.8.1",
     "ioredis": "^1.15.1",
     "is-running": "^2.0.0",
+    "lodash": "^4.13.1",
```
evantahler: never!

initializers/taskMiddleware.js
```diff
@@ -0,0 +1,71 @@
+'use strict';
```
evantahler: Not sure what this file does?

evantahler:
good start!

package.json
l0oky: What's wrong with `lodash`?

initializers/demoMiddleware.js
```diff
@@ -0,0 +1,25 @@
...
+    api.demoMiddleware = {
+      name: "demoMiddleware",
+      priority: 999,
+      preProcessor: function(data, next){
+        api.log("In Preprocessor!", "info");
+        next(null, true);
+      },
+      postProcessor: function(data, next){
+        api.log("In Postprocessor!", "info");
+        api.log("Data keys: ", "info", _.keys(data));
```
S3bb1: you only use lodash for keys ? 

then you can remove the dependency and use Object.keys(data)

gcoonrod: This initializer is only there as a demo. It'll be removed completely.

package.json
gcoonrod: This dep will be removed before the final PR is ready.

initializers/taskMiddleware.js
gcoonrod: This is the resque plugin that handles the middleware processing.

evantahler:
bump! 
Any progress here?

evantahler:
also, please change this PR to merge into the `v14.0.0` branch, not master

evantahler:
I'm going to cut v14 now... which means this isn't going to make it.   We can spend more time on this, and it will make it into v15 then

gcoonrod:
Closing request after talking with @evantahler. I will cut a new PR from the current version in the near future. :pray:

PR #862, opened 2016-06-07
Enable ETAG Headers for Static Files

S3bb1:
This PR enables ETAG Header generation for static served files


servers/web.js
```diff
@@ -105,12 +107,21 @@ var initialize = function(api, options, next){
...
     var headers = connection.rawConnection.responseHeaders;
     if(error){ connection.rawConnection.responseHttpCode = 404; }
     if(ifModifiedSince && lastModified <= ifModifiedSince){ connection.rawConnection.responseHttpCode = 304; }
+    if(api.config.servers.web.enableEtag && fileStream){
+      var fileBuffer = !Buffer.isBuffer(fileStream) ? new Buffer(fileStream.toString(), 'utf8') : fileStream;
+      var fileEtag = etag(fileBuffer, {weak: true});
+      connection.rawConnection.responseHeaders.push(['ETag', fileEtag]);
+      var lastModifiedDate = new Date(lastModified);
+      lastModifiedDate.setMilliseconds(0);
+      if(fresh(reqHeaders, {'etag': fileEtag, 'last-modified':lastModifiedDate})){
```
evantahler: Why do you need a whole package (`fresh`) here?  Aren't you just comparing the requests' `if-none-match` header to the new etag?

servers/web.js
```diff
@@ -105,12 +107,21 @@ var initialize = function(api, options, next){
...
+        connection.rawConnection.responseHttpCode = 304;
```
evantahler: If you do have a header match, I think you are just supposed to respond with 304; and not send the body payload.  I think you still will here

test/core/staticFile.js
```diff
@@ -72,6 +72,27 @@ describe('Core: Static File', function(){
...
+      var etag = response.headers['etag'];
+      var options = {
+        url: url + '/simple.html',
+        headers: {
+          'If-None-Match': etag
+        },
+        method: 'get'
+      };
+      request(options, function(error, response){
+        response.statusCode.should.equal(304);
```
evantahler: test should ensure that there is no body returned

evantahler:
awesome start!

test/core/staticFile.js
S3bb1: fixed it

servers/web.js
S3bb1: No, in the line above it is handled the same. and the check after the logic checks if the responsecode is !== 304 :)

servers/web.js
S3bb1: yeah thats right. I've removed the module and added the check directly

evantahler:
💪

PR #870, opened 2016-06-15
Electron support, avoid bug electron/electron#5454

semiosis:
There's a bug in Electron causing directory listing of an ASAR subdirectory to return a listing of the ASAR root, when the path to list has a trailing slash.  To fix:
- remove trailing path separator from `dir` in recursive directory glob
- use path.join(dir, file) instead of path.normalize(dir + file)


evantahler:
good change! I think inside the ASAR, it looks like your app boots at `/`... which I guess we've never encountered before!

semiosis:
Thank you

PR #871, opened 2016-06-15
Delay log directory creation until after configs are loaded

semiosis:
Support for Actionhero installed on a read-only filesystem.

Actionhero checks to ensure its log directory exists by trying to create the log directory.  It does this during config loading.  As a result, when loading the default config it tries to make a log dir inside the node_modules directory.  If that directory is on a read-only filesystem we crash.

One possible solution (presented here) is to delay the directory creation check until after all configs have been loaded.  This way even if Actionhero is installed on a read-only filesystem, the user can provide an alternate log location on a writable filesystem, and we won't try to create any log directories until all configs have been loaded.

### DIFF
--- a/config/logger.js
+++ b/config/logger.js
@@ -17,18 +17,18 @@ exports['default'] = {
     }
 
     // file logger
-    if(api.config.general.paths.log.length === 1){
-      var logDirectory = api.config.general.paths.log[0];
-      try{
-        fs.mkdirSync(logDirectory);
-      }catch(e){
-        if(e.code !== 'EEXIST'){
-          throw(new Error('Cannot create log directory @ ' + logDirectory));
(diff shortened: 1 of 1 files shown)


evantahler:
First, I love this hack to make a fake transport for winston to delay execution.  However, I think there might be a large problem if the local options for `api.config.general.paths.log` aren't loaded in by now.  

I'll take a look

evantahler:
whoops, I read your change entirely wrong.   Rater than make a new `logger.transports.push`, why not just move the `fs` steps into the `logger.transports.push` block, before the return?

semiosis:
That's what I did 😃

semiosis:
The diff view might make it look weird though

evantahler:
👍

PR #872, opened 2016-06-16
remove vestigial code fragment from comment

muralisr:
comment contains '); which seems to be a remnant of some old code, it is out of place now in the comment


evantahler:
Thanks!

PR #874, opened 2016-06-18
fix function name typo

muralisr:
api.log was mistyped as api.lot in the documentation that shows up on the docs page on the website


evantahler:
👍

PR #883, opened 2016-06-27
Add overwriteConfig parameter to the link function, fix check if config folder not exists

S3bb1:
This PR adds a new parameter "overwriteConfig" to the "link" method for plugins.

When you add the parameter as argument, all config files will be overwritten in the config folder.

This PR also adds a check if a config folder exists in the linked plugin. Before, the link method crashes.


evantahler:
Neat!

PR #892, opened 2016-07-13
Fix for lack of javascript folder under public

krishnaglick:
ActionHero likes to change actionheroClient.js/.min.js so I elected to not track them as they'd be system-specific anyway. This resulted in the javascript folder not existing on other systems, which caused the tests to fail. This is a simple fix to create the javascript folder if it hasn't been.


evantahler:
Thanks!

I'm going to accept this merge, but I'll change back all `const` to `var` (to match the style).  At some point in the future we'll move actionhero up to ES6

PR #895, opened 2016-07-15
Better status action

evantahler:
Add more data

- include event loop delay
- include memory consumption (HEAP)
- total resque queue length

Return a "healthy" or "unhealthy"  response based on this new data (and configurable via ENV)

### DIFF
--- a/actions/status.js
+++ b/actions/status.js
@@ -1,3 +1,11 @@
+var path = require('path');
+var packageJSON = require(path.normalize(__dirname + path.sep + '..' + path.sep  + 'package.json'));
+
+// These values are probably good starting points, but you should expect to tweak them for your application
+var maxEventLoopDelay    = process.env.eventLoopDelay || 5;
+var maxMemoryAlloted     = process.env.maxMemoryAlloted || 200;
+var maxResqueQueueLength = process.env.maxResqueQueueLength || 1000;
+
 exports.status = {
   name: 'status',
   description: 'I will return some basic information about the API',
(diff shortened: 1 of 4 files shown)


actions/status.js
```diff
@@ -9,10 +17,69 @@ exports.status = {
   },
 
   run: function(api, data, next){
+
+    /* --- Define Helpers --- */
+
+    var checkRam = function(callback){
```
gcoonrod: Why not move these helper functions into `api.status` initializer? Provide a couple OOTB status apis.

actions/status.js
```diff
@@ -9,10 +17,69 @@ exports.status = {
...
+
     data.response.id                = api.id;
     data.response.actionheroVersion = api.actionheroVersion;
     data.response.uptime            = new Date().getTime() - api.bootTime;
+    data.response.name              = packageJSON.name;
+    data.response.description       = packageJSON.description;
+    data.response.version           = packageJSON.version;
 
-    next();
+    checkRam(function(error){
```
gcoonrod: No async series?

initializers/utils.js
```diff
@@ -244,6 +245,32 @@ module.exports = {
       return {host: host, port: parseInt(port, 10)};
     };
 
+    api.utils.eventLoopDelay = function(itterations, callback){
+      var intervalJobs = [];
+      var intervalTimes = [];
+
+      var i = 0;
+      while(i < itterations){
```
gcoonrod: Should we do a check on `itterations` to avoid any endless while loops?

actions/status.js
evantahler: I did that at first (and started storing samples), but the more I thought about it, the more I realized that what you consider "important" for health varies drastically between applications.  To that end, I didn't want to add *any* initializers for this, but to showcase *some* of the things you might check, and how to do it.

That'' also why I'm using ENV here and not configuration varialbes.  I expect this file to be drastically changed in a production project.

actions/status.js
evantahler: this needs to work in new projects which won't have async installed by default.

actions/status.js
gcoonrod: True, but I think that AH should have a minimum opinion about whether or not an individual node is healthy. Now, you could achieve that with documentation. But I personally think that adding a few initializers is valuable.

evantahler: I disagree.  For example, the event-loop check seems like a sefe check for a node that is only serving HTTP traffic.  However, if you have a node which is only doing tasks, it's totally fine to block the event loop... and if you are doing something like image processing, that's expected behavior.

initializers/utils.js
evantahler: yep! https://github.com/evantahler/actionhero/pull/895/commits/def9c9d97f1a48fd9a8e2b1decfd9ac06a64e3c9

initializers/utils.js
```diff
@@ -249,6 +249,8 @@ module.exports = {
       var intervalJobs = [];
       var intervalTimes = [];
 
+      if(!itterations){ return callback(new Error('itterations is required')); }
```
gcoonrod: Won't prevent someone passing in `Infinity` and creating an endless loop. But that would be a very specific action.

evantahler: Yeah... That's their own fault then!

PR #896, opened 2016-07-17
[RFC] Task Middleware: Round Two

gcoonrod:
This PR is my second attempt at implementing some basic task middleware that behaves like the action middleware.

Still Todo:
- [x] Tests
- [x] Docs


initializers/taskTimer.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
```
evantahler: This is a a great example for the docs, but probably doesn't belong checked in :D

initializers/tasks.js
```diff
@@ -297,11 +331,34 @@ module.exports = {
...
+        api.utils.recursiveDirectoryGlob(p).forEach(function(f){
+          api.tasks.loadFile(f, reload);
+        });
       });
-    });
+    }
+
+    api.tasks.addMiddleware = function(middleware, callback){
+      if(!middleware.name){ return callback(new Error('middleware.name is required')); }
+      if(!middleware.priority){ middleware.priority = api.config.general.defaultMiddlewarePriority; }
```
evantahler: Please add a note to that config file that `api.config.general.defaultMiddlewarePriority` now effects both tasks and actions.

initializers/tasks.js
```diff
@@ -297,11 +331,34 @@ module.exports = {
...
+      middleware.priority = Number(middleware.priority);
+      api.tasks.middleware[middleware.name] = middleware;
+      if(middleware.global === true){
+        api.tasks.globalMiddleware.push(middleware.name);
+        api.tasks.globalMiddleware.sort(function(a, b){
```
evantahler: Now that both actions and tasks are using this sort, it should probably be moved in to utils.

evantahler:
This is excellent, and on the right track!

evantahler:
You can now change `node-resque` to version `v2.1.0` and use `worker.performInline` in specHelper to help with the tests. 

https://github.com/taskrabbit/node-resque/releases/tag/v2.1.0

gcoonrod:
Hmm, not sure why tests won't pass on Travis...

initializers/specHelper.js
```diff
@@ -161,6 +162,27 @@ module.exports = {
         api.tasks.tasks[taskName].run(api, params, next);
       };
 
+      api.specHelper.runFullTask = function(taskName, params, next){
+        var options = {
+          connection: {},
+          queues: api.config.tasks.queues || ['default']
+        };
+        var worker = new NR.worker(options, api.tasks.jobs);
```
evantahler: I think without a connection, it will try to connect to redis in all cases.  Best to pass in redis object (api.redis.clients.tasks) like we do in `initializers/resque` does

evantahler: This might be the source of your travis trouble

initializers/specHelper.js
```diff
@@ -161,6 +162,27 @@ module.exports = {
...
+        worker.connect(function(error){
+          if(error){
+            return next(error);
+          }
+
+          worker.performInline(taskName, params, function(error, result){
```
evantahler: Just a shorthand note: what you have written is the same as `worker.performInline(taskName, params, next)`

initializers/tasks.js
```diff
@@ -44,13 +46,45 @@ module.exports = {
...
+
+              if(self.worker.queueObject){
+                self.queueObject = self.worker.queueObject;
+              }else{
+                self.queueObject = self.worker;
+              }
+            };
+
+            if(api.tasks.middleware[m].preProcessor){ plugin.prototype.before_perform = api.tasks.middleware[m].preProcessor; }
+            if(api.tasks.middleware[m].postProcessor){ plugin.prototype.after_perform = api.tasks.middleware[m].postProcessor; }
```
evantahler: You are only using 2/4 of the possible node-resque hooks.  After this is merged in, please add a github issue to also allow for this middleware to modify `api.tasks.enqueue` with the `before_enqueue` and `after_enqueue` hooks as well

gcoonrod: Those are not trivial changes. Why do you want to expose them?

evantahler: Completeness?  Doing so is certainly not part of this PR

evantahler:
GREEN!

actions/randomNumber.js
```diff
@@ -4,6 +4,9 @@ exports.randomNumber = {
   outputExample: {
     randomNumber: 0.123
   },
+  inputs: {
```
evantahler: I'm confused about why `test` needs to be hacked into params?  You aren't using an action at all in your work, so why does this mater?  Also, there are better ways of getting this in if you need it than polluting an action's definition.

test/tasks/middleware.js
```diff
@@ -0,0 +1,67 @@
...
+
+  after(function(done){
+    api.tasks.globalMiddleware = [];
+    actionhero.stop(function(){
+      done();
+    });
+  });
+
+  it('can modify parameters before a task and modify result after task completion', function(done){
+    api.specHelper.runFullTask('runAction', taskParams, function(error, response){
```
evantahler: Oh! You are trying to use the 'run action' task.  I think that's an overly complicated task to use... just define a new simple one in-line as part of your test.  See https://github.com/evantahler/actionhero/blob/master/test/core/tasks.js#L18-L78

evantahler:
I think this is good to go! 
Please change the branch you are merging into to the new `v15` branch!

PR #926, opened 2016-08-18
no callback for api.tasks.addMiddleware

evantahler:
It's not needed as there is nothing async here; and this will better match all other addMiddleware signatures.


evantahler:
cc @gcoonrod

PR #930, opened 2016-08-23
Added watchFolderAndAct()

paulxtiseo:
Added watchFolderAndAct() + docs. Not sure if I should have added tests, but could not find tests for the existing watchFileAndAct() anyways.

### DIFF
--- a/initializers/config.js
+++ b/initializers/config.js
@@ -26,6 +26,7 @@ module.exports = {
     // reloading in development mode
 
     api.watchedFiles = [];
+    api.watchedFolders = [];
 
     api.watchFileAndAct = function(file, callback){
       file = path.normalize(file);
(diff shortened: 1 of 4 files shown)


l0oky:
Your pull request seems to fail tests

evantahler:
Yup, looks like tests fail here.  Also:
- Please add a test which demonstrates and "locks in" your new functionality
- This new feature seems to replace some of the code that exists (in action, init, and task loaders).  Please refactor the code to use this new method!

paulxtiseo:
- Where do I put a test for a new API function.
- I believe on the second point I added a second test to an if statement (line 122) to `initializers/utils.js` that may be a problem? If so, I'm not sure how to "refactor the code to use this new method"?

evantahler:
@paulxtiseo Bump?

evantahler:
you can put the test here perhaps https://github.com/evantahler/actionhero/blob/master/test/core/api.js, or just make a new file, that's OK too. 

As far as refactoring, looking though the code again, it's rare that we are watching whole folders (and events happen on the file more than the folder), so maybe there's no refactoring needed.

evantahler:
Closing this due to lack of updates

PR #937, opened 2016-09-06
Fix "connected" event on websocket fired too early

S3bb1:
When you listen on the "connected" event on the websocket connect and you fire immediately a action over the client lib the "actionWeb" function is used instead of the "actionWebSocket" function because ` self.state = 'connected';` is set after the "connected" event is emitted.

This fix solves the problem


l0oky:
Wait, does It matter If you place that line above or below `If statement` when nodejs runs this non-blocking?

S3bb1:
Try it:

change your index.html (at the bottom) to:

``` javascript
    var client = new ActionheroClient();
    client.connect();
    client.on('connected', function(){
      client.action('showDocumentation', function(data){
        if(data.error){ window.alert(data.error); }
        populateDocumentation(data)
      });

      client.action('status', function(data){
        if(data.error){ window.alert(data.error); }
        populateStatus(data)
      });
    });
```

this **should** wait for the connected event from the websocket. And see the response : 

image


As you can see, the actions are called via actionWeb (REST) instead of websockets.

When you change the line to my pull request the network tab looks like:
image

evantahler:
great catch and explanation! 

I think writing a test for this behavior is a bitt *to* tricky, so we can merge this as-is.

Thanks!

l0oky:
Pardon me, I didn't notice this is from `public/javascript/actionheroClient.js`. Good catch!

PR #938, opened 2016-09-07
Fix ETAG issue for files

S3bb1:
We've found a issue with the etags, every file had the same etag. This was because fileStream.toString() is [object Object] and not the content istself. So every ETAG was generated for [object Object]

I've added a test to check if the ETAG is NOT the same for different files.


evantahler:
great find!

However, don't use a `sync` fall to the filesystem, as this will block all other operations.

S3bb1:
yeah, good point. I will adjust this to be async

evantahler:
bump?

S3bb1:
Sorry for the delay, was a little busy the last days. I will push the changes today.

evantahler:
Yay!

PR #939, opened 2016-09-07
add workerTitlePrefix option

ericli1018:
run cluster mode, it would be useful to show who joined cluster.

### DIFF
--- a/bin/methods/start/cluster.js
+++ b/bin/methods/start/cluster.js
@@ -36,6 +36,7 @@ var optimist  = require('optimist');
 var argv = optimist
   .describe('workers', 'How many worker node processes')
   .default('workers', os.cpus().length)
+  .describe('workerTitlePrefix', 'Worker title prefix, set `--workerTitlePrefix=hostname`, your app.id would be like `your_host_name-#`')
   .argv;
 
 /////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


bin/methods/start/cluster.js
```diff
@@ -146,6 +147,12 @@ var ActionHeroCluster = function(args){
 };
 
 ActionHeroCluster.prototype.defaults = function(){
+  var title = optimist.argv.workerTitlePrefix;
```
evantahler: Don't set defaults within the app, but use the `.default()` directive within optimist (see line 38 as an example)

bin/methods/start/cluster.js
```diff
@@ -36,6 +36,7 @@ var optimist  = require('optimist');
 var argv = optimist
   .describe('workers', 'How many worker node processes')
   .default('workers', os.cpus().length)
+  .describe('workerTitlePrefix', 'Worker title prefix, set `--workerTitlePrefix=hostname`, your app.id would be like `your_host_name-#`')
```
evantahler: this description is quite long.  Shorten it here and add more detail in the help text and documentation site.

evantahler:
A good feature addition! Please address the comments and note you change in the help.txt and website documentation:

- https://github.com/evantahler/actionhero/blob/master/bin/templates/help.txt
- https://github.com/evantahler/actionhero/blob/ec9ea30f15aa3ce0bb1dec377038e7bd94ff5dd2/site/source/includes/docs/ops/running_actionhero.md

evantahler:
👍

PR #941, opened 2016-09-14
fix no workerTitlePrefix then show NaN title

ericli1018:
fix a little bug.

I had assigned `.default('workerTitlePrefix', 'actionhero-worker-')`, but look no use.
So I add some code to fix that issue.

### DIFF
--- a/bin/methods/start/cluster.js
+++ b/bin/methods/start/cluster.js
@@ -404,9 +404,9 @@ module.exports = function(api){
 
       var title = optimist.argv.workerTitlePrefix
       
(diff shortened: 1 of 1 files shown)


evantahler:
please use `===`

bin/methods/start/cluster.js
```diff
@@ -404,9 +404,9 @@ module.exports = function(api){
 
       var title = optimist.argv.workerTitlePrefix
       
-      if(title == null || title == '')
```
l0oky: Missing ===

PR #942, opened 2016-09-15
Updates error string formatting

dangerdespain:
I misnamed a plugin reference, and an error had me temporarily confused.

"cannot find linked refrence toah-sequelize.plugin.link "

This fix would format the error as -
"cannot find linked refrence to ah-sequelize.plugin.link"


evantahler:
💭

PR #949, opened 2016-09-25
Respect accept-encoding priority order as set by request header

aurasalexander:
I have stumbled into this pull request while debugging a problem with IE and activated actionhero compression.
IE will fail to display contents when compression and http secure is activated at the same time. I searched a bit and found some sources that involved headers bytes of the DEFLATE encoding being set or not resulting in IE being able to decode the stream (or fail to).
In succession, i tried switching the order of the deflate/gzip accept-encoding check to see how well IE is able to handle GZIP instead of the previously favored DEFLATE. ...And it did handle it well. I read that IE will never send the DEFLATE argument as first item in the list of supported encodings. Therefore I implemented a check about the order of the accepted encodings when applying a compression.

This won't fix the underlying problem, but the root of this bug seems not to be directly in actionhero.
The correct handling of the encoding priority should be done anyway and happens to fix the main symptom of the compression+https IE bug.

### DIFF
--- a/servers/web.js
+++ b/servers/web.js
@@ -175,16 +175,19 @@ const initialize = function(api, options, next){
     // https://nodejs.org/api/zlib.html#zlib_zlib_createinflate_options
     // See http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.3
     if(api.config.servers.web.compress === true){
-      if(acceptEncoding.match(/\bdeflate\b/)){
-        headers.push(['Content-Encoding', 'deflate']);
-        compressor = zlib.createDeflate();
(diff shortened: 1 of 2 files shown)


evantahler:
Great find, and great fix! 

Could you please include a test for the web server to ensure that header priority is maintained?  Request, which we use in our test suite, allows fairly detailed controls over headers: https://github.com/request/request#custom-http-headers

aurasalexander:
Sure, I will take a look into it some hours from now.

evantahler:
Comments in Slack https://actionherojs.slack.com/files/evantahler/F2G6JMKCZ/Untitled.js

PR #950, opened 2016-09-25
Fix cache headers

aurasalexander:
With this push, no expires or last-modified headers are being sent when etags are enabled. Last-modified and etags are both 'weak' caching headers and we should opt for one; if the developer enabled etags we should make sure they are being used - but atm we let the clients browser implementation decide.

Additionally expires headers won't even be set with etags disabled, as the cache-control header supersedes the 'outdated' expires header - it is replaced by cache-control; both are 'strong' caching headers, we should opt for the newer one, even if it isn't new at all by now.

The current implementation wasn't working right, especially on mobile devices, with files getting cached longer than they should or not at all, depending on the platform.

Caching was a bit of a mess, i refactored the sendFile flow a bit and repaired if-modified-since requests. When a browser receives a page with a last-modified header, it may send the date received with the if-modified-since header, expecting a 304 if it didn't change. But nearly all of these requests will result in the file being sent even if unchanged as the milliseconds of the 'lastModified' (mTime) may differ from zero while the time string from browser (bTime) doesn't contain milliseconds and are therefor zero. Hence mTime <= bTime is false nearly every time.

For more information you can look into the [caching guide from google](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching?hl=en).

### DIFF
--- a/config/servers/web.js
+++ b/config/servers/web.js
@@ -37,8 +37,12 @@ exports['default'] = {
         simpleRouting : true,
         // queryRouting allows an action to be defined via a URL param, ie: /api?action=:action
         queryRouting : true,
-        // The header which will be returned for all flat file served from /public; defined in seconds
+        // The cache or (if etags are enabled) next-revalidation time to be returned for all flat files served from /public; defined in seconds
         flatFileCacheDuration : 60,
+        // Add an etag header to requested flat files which acts as fingerprint that changes when the file is updated;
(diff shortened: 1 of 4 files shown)


evantahler:
related to https://github.com/evantahler/actionhero/pull/938 cc @S3bb1

evantahler:
This is wonderful.  Thank you!

evantahler:
@aurasalexander please join us on Slack if you have any questions/comments for us! http://slack.actionherojs.com/

PR #953, opened 2016-09-30
we use sendfile to send buffers, not files. This additional check will allow for buffer streams to be sent via sendfile too

bluesunrise:


evantahler:
Cool.  Can you please add a test that demonstrates this functionality (and locks it in?).

PR #961, opened 2016-10-12
add `actionhero version` + allow `help` and `version` to be run outside of actionhero projects

evantahler:
1) add `actionhero version` to return the version of ActionHero being used
2) When using a globally-installed version of actionhero, allow the following subset of commands to be used outside of an ActionHero project directory:
- `actionhero help`
- `actionhero generate`
- `actionhero version`

Other commands will remain crashing outside of an ActionHero project folder, as they will require config files.


evantahler:
solves https://github.com/evantahler/actionhero/issues/960

Issue #962, opened 2016-10-18
lastModified.setMilliseconds error occurring nightly


BigForNothing:
# lastModified.setMilliseconds error occurring nightly
Every night for awhile ( I think it may have started after version 15 ), I've started getting an error in the logs. It happens regardless if it's in a cluster or not. I've added some console.log outputs to node_modules/actionhero/servers/web.js (after line 130, inside the if statement to help. The if statement now looks like this: 

    if(reqHeaders['if-modified-since']){
      console.log('reqHeaders', reqHeaders);
      console.log(reqHeaders['if-modified-since']);
      ifModifiedSince = new Date(reqHeaders['if-modified-since']);
      lastModified.setMilliseconds(0);
      if(lastModified <= ifModifiedSince){connection.rawConnection.responseHttpCode = 304; }
      return sendRequestResult();
    }

The error: TypeError: Cannot read property 'setMilliseconds' of undefined
- ActionHero Version: `15.1.0`
- Node.js Version: `6.3.1`
- Operating System `Ubuntu 14.04 Trusty`

# Steps to reproduce your error
- Start actionhero. Normal or cluster mode doesn't matter.
- No other activity is happening until sometime in the evening.

Stack trace, including added console.log output:

```
reqHeaders { host: 'www.baidu.com',
  'user-agent': 'Mozilla',
  'if-modified-since': 'Thu, 19 Apr 2012 09:51:20 GMT' }
Thu, 19 Apr 2012 09:51:20 GMT
/home/ubuntu/xxxxx/node_modules/actionhero/servers/web.js:134
      lastModified.setMilliseconds(0);
                  ^

TypeError: Cannot read property 'setMilliseconds' of undefined
    at server.sendFile (/home/ubuntu/xxxxx/node_modules/actionhero/servers/web.js:134:19)
    at api.staticFile.get (/home/ubuntu/xxxxx/node_modules/actionhero/initializers/genericServer.js:99:14)
    at Object.api.staticFile.sendFileNotFound (/home/ubuntu/xxxxx/node_modules/actionhero/initializers/staticFile.js:80:9)
    at Object.api.staticFile.get (/home/ubuntu/xxxxx/node_modules/actionhero/initializers/staticFile.js:29:16)
    at checkExistence (/home/ubuntu/xxxxx/node_modules/actionhero/initializers/staticFile.js:45:32)
    at fs.stat (/home/ubuntu/xxxxx/node_modules/actionhero/initializers/staticFile.js:86:13)
[...]

PR #963, opened 2016-10-19
Addressing if-modified-since issue on a 404

BigForNothing:
This should resolve the issue for #962. 

I've included a test to reproduce, and within servers/web.js simply moved the if(error) statement above the if-modified-since if statement.

The issue seems to be a 404 file, with a If-Modified-Since header set. I originally thought it might be related to a POST method but it was directly related to the header being set. I just double checked GET and it works as expected. Because of that, in this pull request, you'll see the test using the POST method.

### DIFF
--- a/test/servers/web.js
+++ b/test/servers/web.js
@@ -25,14 +25,14 @@ describe('Server: Web', function(){
   });
 
   it('file: 404 pages from POST with if-modified-since header', function(done) {
-    let file = Math.random().toString(36);
-    let options = {
+    var file = Math.random().toString(36);
+    var options = {
       url: url + '/' + file,
       headers: {
         'if-modified-since': 'Thu, 19 Apr 2012 09:51:20 GMT'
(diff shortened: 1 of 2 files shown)


test/servers/web.js
```diff
@@ -24,6 +24,22 @@ describe('Server: Web', function(){
     });
   });
 
+  it('file: 404 pages from POST with if-modified-since header', function(done) {
```
evantahler: looks like you need to add a `use strict` to the top of this file so ES6 declarations work in node v4

evantahler:
This is a good fix!  Please update the test so it passes, and I'll merge it in!

test/servers/web.js
BigForNothing: Will do, do you want me to change it to a request.get instead of request.post as well?

evantahler:
Thanks!

PR #971, opened 2016-10-27
Prevent xss in file server

submitteddenied:
This PR closes a vulnerability in the file server where an attacker can request a file name which includes HTML characters that are not escaped in the response.


evantahler:
see https://github.com/evantahler/actionhero/pull/972

PR #976, opened 2016-10-30
Do not localize log messages by default; add option to do so

evantahler:
solves https://github.com/evantahler/actionhero/issues/974

Creates a new config option `api.config.logger.localizeLogMessages` which allows developers to opt-into localizing all system log messages (`api.log`).  

The default changes in this pull request from the previous "enabled implicitly" to "disabled explicitly"


gcoonrod:
I like this approach. 👍

PR #997, opened 2017-01-05
Change initializer identifier to file path instead of file name

S3bb1:
This small change changes the identifier for the initializer keys at startup to the full filepath instead of the filename.

This prevents errors when modules include the same initializer filenames.


S3bb1:
i agree with your argumentation. So maybe we can add your suggested logic for initializers which came from the actionHero itself (actions,actionProcessor,...) and use the path identifier for others ?

evantahler:
I would like to have consistent logic all the way though wherever possible... ActionHero uses ActionHero to be ActionHero after all!

S3bb1:
okay so i've removed the file path identifier and included a error message if the initializer already exists

evantahler:
Interesting! You have chosen to allow the application to boot, but "skipped" any initializers which duplicate a name.  

I personally would prefer the application to crash with the message  `Initializer %s{} already exists`.  The way you have it here would boot the application into a state which is *unexpected* for the developer.   A crash is a clear signal that something is wrong, and should be fixed in development.  

A crash like this wouldn't make it to production, so I don't think we need to be safe about it.

S3bb1:
mean't like this ?

actionhero.js
```diff
@@ -115,6 +115,7 @@ actionhero.prototype.initialize = function (params, callback) {
 
   this.configInitializers.push(() => {
     let customInitializers = []
+    let ignoredInitializers = []
```
evantahler: This should no longer be called `ignoredInitializers`, perhaps `duplicatedInitializers`?

actionhero.js
```diff
@@ -132,8 +133,15 @@ actionhero.prototype.initialize = function (params, callback) {
       let fileParts = file.split('.')
       let ext = fileParts[(fileParts.length - 1)]
       if (ext === 'js') {
-        delete require.cache[require.resolve(file)]
-        this.initializers[initializer] = require(file)
+        // check if initializer already exists (exclude utils and config)
+        if (this.initializers[initializer] &&
+           file !== path.resolve(__dirname, 'initializers', 'utils.js') &&
+           file !== path.resolve(__dirname, 'initializers', 'config.js')) {
+          ignoredInitializers.push(file)
```
evantahler: Why have this new array at all?  why not just fail here directly with a `throw`?

actionhero.js
S3bb1: fully agree with this :)

actionhero.js
S3bb1: yeah if you have more than one conflicting initializer this would be helpful to know all, not just the first one... so you would fix the first initializer and actionhero keeps crashing again

evantahler:
Makes sense! Merged!

PR #998, opened 2017-01-18
Do not require a fileLength when sending a stream

jacekgdudek:


jacekgdudek:
Note: Required when f.e. making a call to elasticsearch during the action call. Acquiring elasticsearch readStream and passing it as a sendFile using ```api.servers.servers.web.sendFile```. At that point we don't know how big (compressed or not) the stream is going to be.

evantahler:
This is a good change.  Can you add in a test that shows it is working as expected?  You may be able to extend this test https://github.com/evantahler/actionhero/blob/master/test/actions/sendBuffer.js

servers/web.js
```diff
@@ -209,7 +209,10 @@ const initialize = function (api, options, next) {
         connection.rawConnection.res.writeHead(responseHttpCode, headers)
         fileStream.pipe(compressor).pipe(connection.rawConnection.res)
       } else {
-        headers.push(['Content-Length', fileLength])
+        // file length might be null if we don't know how long the stream is
+        if (fileLength) {
```
gcoonrod: Please add a test for this change as per @evantahler request.

evantahler:
Bump! 
any progress here?

servers/web.js
jacekgdudek: Added

jacekgdudek:
Sorry for the delay. Hope the test is up to your satisfaction. Is there any chance to have it also re-released under v15.0.3?

evantahler:
Thanks! 👍

evantahler:
re: v15
We've never back-ported a fix to an older major version, sorry.  This will be released in the v16.x  world.  The migration from v15 to v16 is quite small https://github.com/evantahler/actionhero/releases

jacekgdudek:
Thanks!

PR #1010, opened 2017-03-01
newly generated projects include Standard.js and pass lint

evantahler:


bin/templates/task.js
```diff
@@ -9,6 +9,9 @@ exports.task = {
 
   run: function (api, params, next) {
     // your logic here
-    next (error, resultLogMessage)
+    let error = new Error('something has gone wrong')
```
gcoonrod: By default the generated tasks will fail?

evantahler: Good point.  I'll put these in comments to ensure that it passes

evantahler:

PR #1022, opened 2017-03-13
fix(http): use documented api methods to set headers

gergelyke:


evantahler:
Thoughts on the direct implantation: 
- Why not remove `cleanHeaders` here as well?
- `cleanHeaders` reverses the header array and the iterates to ensure that the "newest" header added if there are duplicates will win out.  You are accomplishing  this directly with hash-propery uniqueness. Nice!

This seems like a good step to address your specific goal in using the res.writeHead API as directly as we can.  This will not address your or @neilstuartcraig desire to use `data.connection.rawConnection.setHeader('myHeader', 'cool')` directly.  I actually really like that approach.  This as written so far is not a breaking change... feel free to make it one!  My only concern is how to "merge" duplicate cookie headers, but you can create a helper for that.

evantahler:
Lets merge this in as-it-is, and I'll start a new issue to refactor how we deal with headers in general.

evantahler:
https://github.com/actionhero/actionhero/issues/1025

evantahler:

PR #1026, opened 2017-03-24
add schema inputs for Actions

witem:
Add schema input type for actions:
```
inputs: {
  userData: {
    required: false,
    schema: {
      id: {
        required: false,
        validator: validator('isInt', 'id validate error'),
      },
      name: {
        required: false,
        validator: validator('isLength', 'name validate error'),
      },
    },
  },
},
```

- [x] add tests
- [x] run `npm test`

### DIFF
--- a/test/core/api.js
+++ b/test/core/api.js
@@ -192,6 +192,7 @@ describe('Core: API', () => {
     after((done) => {
       delete api.actions.actions.testAction
       delete api.actions.versions.testAction
+      api.config.general.missingParamChecks = [null, '', undefined]
       done()
     })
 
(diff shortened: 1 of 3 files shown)


evantahler:
Can you describe this PR? What are your goals?  Can you share a use-case?  Is this a breaking change?

crrobinson14:
I went partially down this road myself at one point but I realized that it would probably be a lot of work to do "really well", which is to say, dealing properly with array/nested object types and things like that.

I would suggest you re-approach this with a simple initializer that exposes a validation call for the top level element itself, and uses http://json-schema.org/ and a validator like https://github.com/epoberezkin/ajv to do the checking. It would be pretty easy to do, would rely on standards everybody else is doing, and would make a great plugin without adding complexity to AH Core. I'd be happy to help if necessary.

witem:
@evantahler 
This PR add the ability to handle nested objects in inputs. Any nesting.
They support the same parameters as regular inputs(default, formatter, validator, required).

Goals: accept nested objects.
Use-case: need accept, validate and formatter request with nested object to API. Example:
```
exports.addUser = {
  name: 'api/addUser',
  description: 'I add user',
  
  firstName: { required: true },
  lastName: { required: false },
  username: { required: true },
  address: {
    required: false,
    schema: {
      country: {
        required: true,
        default: 'USA'
      },
      state: { required: false },
      city: {
        required: true,
        formatter: (val) => `City:${val}`,
        validator: (val) => val.length > 10,
      }
    }
  }
  run: () => {},
}
```

No, breaking change.
  
@crrobinson14 it`s hard to realize as plugin with default, validator, formatter and required/

witem:
@crrobinson14 Plugin that overwrites the standard function, not very good solutions.

test/core/api.js
```diff
@@ -263,6 +264,114 @@ describe('Core: API', () => {
...
+        api.specHelper.runAction('testAction', {schemaParam: {requiredParam: []}}, (response) => {
+          expect(response.params.schemaParam.requiredParam).to.have.length(0)
+          done()
+        })
+      })
+    })
+
+    it('will fail for missing or empty string params', (done) => {
+      api.specHelper.runAction('testAction', {schemaParam: {requiredParam: ''}}, (response) => {
+        expect(response.error).to.contain('required parameter for this action')
```
evantahler: Can you be more explicit in this test?  Do you expect the error to be:
- `schemaParam.requiredParam  is a required parameter for this action`
- `schemaParam[requiredParam]  is a required parameter for this action`
- `requiredParam  is a required parameter for this action`

etc?

evantahler:
@witem your work here looks compete to me, with good test coverage! 

@crrobinson14 can you share what trouble you ran into regarding `dealing properly with array/nested object types`?

test/core/api.js
witem: expect: `schemaParam.requiredParam is a required parameter for this action`

evantahler: Awesome.  Please update the test to check for exactly that string.

witem: done

crrobinson14:
@evantahler Sorry, to clarify, I wasn't taking issue with the PR. Just suggesting an alternate route I've used that's fairly easy to implement with an initializer. It meshes well with the Swagger typed-object approach where to define the schema for an object ahead of time and make it part of your API spec. For implementation in an action, you can do things like:

```javascript
bookData: {
    required: false,
    validateType: 'Book'
  },
```

The initializer can look for a `validateType` (or similar) key defined in this way, and call a JSON schema validator tor check this.

The reason this is valuable is when you have multiple actions all needing similar/identical type checks. Defining the validations purely in the action template could be messy in those cases. But obviously it's a valid approach for some others - that's why I call this "another option", not "it should be done like..."

evantahler:
I like @witem's approach here very much.  It's non-breaking, and it follows the same patterns and conventions we already have.  In fact, this is probably what people *expect* should happen anyway with these nested uploads, as I've seen questions about it before. 

@witem I think this looks great. The only thing missing now is to update the Docs to talk about this!  At the very least, a section (here)[https://www.actionherojs.com/docs/core/#actions-inputs] is required.  That would be edited here: https://github.com/actionhero/actionhero/blob/master/docs/_includes/docs/core/actions.html

While a bit out of scope for this PR, @crrobinson14 the pattern I generally advise people to follow with shared validtors/formatters is to put them in a separate file and require them when needed, ie: 

```js
const Validators = require('../support/validators.js')

exports.myAction = {
  name: 'myAction',
  description: 'myAction',

  inputs: {
    email: {
      required: true,
      validator: Validators.email,
    },
    password: {
      required: true,
      validator: Validators.password,
    },
  }

  run: function (api, data, next) {
    //... 
    next(null)
  }

}
```

witem:
@evantahler I update docs and add example

witem:
And about validator. I use wrapper on https://www.npmjs.com/package/validator

validatorWrapper.js:
```
const validator = require('validator');
const validationWrapper = function(method, _message, secondArg) {
    let message = _message;
    if (validator[method] == null) {
        throw new Error(`Not exists method for validator: ${method}`);
    }
    if (message == null) {
        message = `validation error with method ${method}`;
    }

    return function(value) {
        try {
            if (validator[method](`${value}`)) {
                return true;
            }
        } catch (err) {
            message = err.message;
        }
        return new this.AppError(message, 100);
    };
};

module.exports = validationWrapper;
```

And in actions in look like:
```
const validator = require('./validationWrapper');


inputs: {
  userEmail: {
    required: true,
    validator: validator('isEmail', 'userEmail validate error'),
  },
},
```

evantahler:
Awesome! Thanks so much!

evantahler:

PR #1031, opened 2017-03-30
Fix cache lock

vladbalmos:
The current implementation of the `cache.lock` method does not check the result of the `setnx` function, which returns either 1 if the key didn't previous exist, and 0 otherwise, this can lead to multiple nodes aquiring the same lock multiple times.

### DIFF
--- a/initializers/cache.js
+++ b/initializers/cache.js
@@ -256,7 +256,7 @@ module.exports = {
               return callback(error)
             } else {
               if (!result) { // value was already set, so we cannot obtain the lock
-                return callback(null, false);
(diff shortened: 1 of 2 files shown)


gcoonrod:
@vladbalmos Can you add a test for this case?

vladbalmos:
Done

gcoonrod:
@evantahler Any comments before I merge this?

evantahler:
Looks great! Thank you for finding this!

vladbalmos:
I'm glad I could help!

PR #1036, opened 2017-04-08
Saner localization

evantahler:
- clarify the usage of i18n and `connection.localize` by converting all localizations to use mustache-style strings.  IE: `data.connection.localize(['hello {{name}}', {name: 'Evan'}])` rather than `data.connection.localize(['hello %s, evan])`
- [BREAKING CAHNGE] Stop localizing all internal and logging messages
  - this feature was rarely used and added complexity.  This also slowed down the logging of message from the server as it required 2 interpolation passes each time
  - convert all internal message interpolation into es6 interpolation, ie: `a log ${message}`.  Now that we required node v4+, this is OK
- [BREAKTING CHANGE] move all user-visible strings from actionhero into the locale file.  This creates a single file which can be used to modify ActionHero strings
  - All actionhero strings reside within an `actionhero` translation namespace: https://github.com/actionhero/actionhero/blob/mustache_i18n/locales/en.json
  - `locales/en.json` is now checked into the project
  - ensure this locale file is copied into a new project via `actionhero generate`
  - `welcomeMessage` and `goodbyeMessage` are removed from config into locale
- ensure that by default, calling either `api.i18n.localize` or `connection.localize` on an unknown string will not cause an error, and this string will be added to the locale file properly

### DIFF
--- a/actions/status.js
+++ b/actions/status.js
@@ -26,7 +26,7 @@ exports.status = {
       data.response.consumedMemoryMB = consumedMemoryMB
       if (consumedMemoryMB > maxMemoryAlloted) {
         data.response.nodeStatus = data.connection.localize('Unhealthy')
-        data.response.problems.push(data.connection.localize('Using more than ' + maxMemoryAlloted + 'MB of RAM/HEAP'))
+        data.response.problems.push(data.connection.localize(['Using more than {{maxMemoryAlloted}} MB of RAM/HEAP', {maxMemoryAlloted: maxMemoryAlloted}]))
       }
 
       callback()
@@ -38,7 +38,7 @@ exports.status = {
         data.response.eventLoopDelay = eventLoopDelay
         if (eventLoopDelay > maxEventLoopDelay) {
           data.response.nodeStatus = data.connection.localize('Node Unhealthy')
-          data.response.problems.push(data.connection.localize('EventLoop Blocked for more than ' + maxEventLoopDelay + 'ms'))
+          data.response.problems.push(data.connection.localize(['EventLoop Blocked for more than {{maxEventLoopDelay}} ms', {maxEventLoopDelay: maxEventLoopDelay}]))
         }
 
         callback()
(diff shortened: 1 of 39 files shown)


actions/status.js
```diff
@@ -26,7 +26,7 @@ exports.status = {
       data.response.consumedMemoryMB = consumedMemoryMB
       if (consumedMemoryMB > maxMemoryAlloted) {
         data.response.nodeStatus = data.connection.localize('Unhealthy')
-        data.response.problems.push(data.connection.localize('Using more than ' + maxMemoryAlloted + 'MB of RAM/HEAP'))
+        data.response.problems.push(data.connection.localize(['Using more than {{maxMemoryAlloted}} MB of RAM/HEAP', {maxMemoryAlloted: maxMemoryAlloted}]))
```
gcoonrod: Is there a reason we can't use ES6 string templates here?

evantahler: Ah, because this is a string we actually want localized, as this message would be viewed by someone "outside" of the server (someone looking at a webpage which hits this API).  You might want to say this sentence in Spanish, so your `/locale/es.json` would have `Uso de más de {{maxMemoryAlloted}} MB de RAM / HEAP`

evantahler: Or, more realistically the within the action would be changed to `data.connection.localize('actions.status.problems. maxMemoryAlloted')` and then each of the locale files would have something like 

```json
{
  "actions": {
    "status": {
      "maxMemoryAlloted": "Uso de más de {{maxMemoryAlloted}} MB de RAM / HEAP"
    }
  }
}
```

gcoonrod: So the actual limitation is that the localized string templates are stored in JSON files which do not support ES6 string templates?

evantahler: I don't think I follow.  i18n is a spec (used by ruby, python, node, etc).  Here's the node library we use https://github.com/mashpie/i18n-node#mustache-support.  Mustache is the generally agreed upon templating library across all implementations.  In general, this style allows you to re-use locale files across many instances of your project, perhaps in your ios app and your server

gcoonrod: My point was that it would be nice to only have a single type of templated string in the codebase. But I understand the limitations of use `i18n-node`.

PR #1038, opened 2017-04-11
Custom CLI Commands

evantahler:
Allow actionhero developers to create new files in `./bin` which can be run via the CLI.  These commands will have access to a the ActionHero `api` and CLI arguments object within a `run` method.

For example, a file in `./bin/redis/keys` would be run via `./node_modules/.bin/actionhero redis keys --prefix actionhero`
```js
'use strict'

module.exports = {
  name: 'redis keys',
  description: 'I list all the keys in redis',
  example: 'actionhero keys --prefix actionhero',

  inputs: {
    prefix: {
      requried: true,
      default: 'actionhero',
      note: 'the redis prefix for searching keys'
    }
  },

  run: function (api, data, next) {
    api.redis.clients.client.keys(data.params.prefix, (error, keys) => {
      if (error) { throw error }

      api.log(`Found ${keys.length} keys:`)
      keys.forEach((k) => { api.log(k) })

      return next(null, true)
    })
  }
}
```

- [breaking change]: ActionHero projects will now be generated with a `./bin` directory.  
- [breaking change]: `./config/api.js` now defines `api.config.general.paths.cli`, which defaults to `path.join(__dirname, '/../bin')`.  This is where the actionhero script runner sources its files.

---

ActionHero CLI commands have:
- name
- description
- example

Inputs for CLI commands have
- required (true/false)
- default (string only)
- note

These are sourced by `actionhero help`, and the example above would return:
```
* redis keys
  description: I list all the keys in redis
  example: actionhero keys --prefix actionhero
  inputs:
    [prefix] (optional)
      note: the redis prefix for searching keys
      default: actionhero
```

This PR includes a refactor of all CLI commands


pckhoi:
A way to configure the commands to also run `start` and `stop` functions of initializers would be nice. Perhaps `serverLevel: "started"` or `serverPhase: "started"` (default to `"initialized"`).

evantahler:
@pckhoi thinking about this more, it is actually dangerous to start the whole server via the CLI.  And, if you are on the same host which you already have a web-server for example, `start` will fail, as the ports for the web are already open and will block.

To that end, if you really do need to `start` something, whatever start does should be exposed at another method on the api object you can call within your method.

PR #1045, opened 2017-04-25
Catch Exceptions in Initializers

S3bb1:
When you create an initializer which throws an Error or Uncaught Exception like:

```
'use strict'

module.exports = {
  loadPriority: 900,
  startPriority: 900,
  initialize: function (api, next) {
    a=b;
    next()
  },

  start: function (api, next) {
    next()
  }
}
```

and a second initializer which runs before the top initializer which try/catches errors in the current initializer like:

```
'use strict'

module.exports = {
  loadPriority: 800,
  startPriority: 800,
  initialize: function (api, next) {
    try{
        next()
    } catch(e) {
        console.dir(e);
        next();
    }
  },

  start: function (api, next) {
    next()
  }
}

```

and run actionHero, the error is thrown but catched by the other initializer and runs in an error:

> if (fn === null) throw new Error("Callback was already called.");
> Error: Callback was already called.

and you can't get in which initializer the error occured.

This PR fixes this issue and catches errors within initializers and prints out the error on the console:

> @ 2017-04-25T18:59:46.782Z - debug: Exception occured in initializer: test  during load ReferenceError: b is not defined


evantahler:
Awesome!

evantahler:
FYI, I've changed the log-level of your error catcher to 'warn', because... that seems important.

2e69814a0663a4c6268ac527b5e03f69a3b6969b

PR #1053, opened 2017-05-15
Proxy setHeader from rawConnection

gcoonrod:
This PR proxies the `setHeader` function from a http response object so that actions running on the web server can use `data.connection.setHeader` instead of `data.connection.rawConnection.res.setHeader`.


evantahler:
I think your tests are good, but the method declaration belongs as part of the web server, not the generic case.  Then, you don't need the conditional. 

Perhaps what we should do is allow `server.buildConnection()` to accept `customMethods` which can be passed in.  Then, in the generic server, `api.GenericServer.prototype.buildConnection` takes all the custom methods and applies them to the connection as it is built.  This way, we can do some checking to ensure that we aren't overriding an existing method/property in a common way.

gcoonrod:
@evantahler Ready for round two!

evantahler:
I like it!
Now please add a Pull request to https://github.com/actionhero/www.actionherojs.com with the documentation update.

evantahler:
Oh, and also the tests should pass :D

gcoonrod:
Sorry this is taking so long. Hope to have time in the coming week to work on it!

servers/web.js
```diff
@@ -342,7 +342,10 @@ const initialize = function (api, options, next) {
         id: fingerprint + '-' + uuid.v4(),
         fingerprint: fingerprint,
         remoteAddress: remoteIP,
-        remotePort: remotePort
+        remotePort: remotePort,
+        customMethods: function (connection) {
+          connection.setHeader = res.setHeader.bind(res)
```
evantahler: shouldn't this be `connection.rawConnection.res.setHeader.bind(res)`

PR #1056, opened 2017-05-17
Allow loading config files from custom plugin folders.

SuatEyrice:
Allow loading config files from custom plugin folders with *.link files.


evantahler:
Hi! Can you explain a little more about your use-case?  We've generally not allowed config files to be read from linked plugins, as they can't be changed.  That is why the act of running `actionhero link` actually copies config files out of your plugin into your main project.  I'd like to hear more about why that behavior doesn't work for you.

Thanks!

PR #1062, opened 2017-05-26
dot-prop package used with api.config.general.filteredParams

chimmelb:
Uses the [dot-prop package](https://github.com/sindresorhus/dot-prop) to allow for filtering parameters using a dot notation.

Added to `api.utils.filterObjectForLogging` since I wanted the filter for my own logging messages as well (not just action logs)

```
//example, as filled from config/api.js
filteredParams: [ 'p4', 'o1.p3' ],

//example, as used in initializers/actionProcessor.js
var params = { p1 : 1, o1 : { p2: 'ok', p3: 'secret' }, p4: 'long string that clogs logs' }
var filteredParams = api.utils.filterObjectForLogging( params)
//filteredParams --> { p1 : 1, o1 : { p2: 'ok', p3: '[FILTERED]' }, p4: '[FILTERED]' }
```


chimmelb:
Not a huge concern, but `api.utils.objClone`, which is used when creating the `filteredParams` object, is not a deep clone. So after logging and filtering a property 2+ levels down, later users of that `data.params` object would find the property filtered as well. Just something I noticed - unsure the need for a "deep clone" here or not.

Also, I see Node4 didn't like my test, I'll fix that.

chimmelb:
Would this be considered ready for release? I'll fix something if that is holding it back.

A separate point release is not necessary for just this PR, just want to be sure it's ready for the next one.

initializers/utils.js
```diff
@@ -350,6 +351,25 @@ module.exports = {
...
+      for (let i in params) {
+        if (api.utils.isPlainObject(params[i])) {
+          filteredParams[i] = api.utils.objClone(params[i])
+        } else if (typeof params[i] === 'string') {
+          filteredParams[i] = params[i].substring(0, api.config.logger.maxLogStringLength)
+        } else {
+          filteredParams[i] = params[i]
+        }
+      }
+      api.config.general.filteredParams.forEach((param) => {
```
evantahler: I think this section only makes sense in the case where the param in question is an object.  I think you can move this under the `if (api.utils.isPlainObject(params[i])) {` check.  

I'm sensitive to not introducing a second loop for every log line

chimmelb: The first loop is the params of the action. This second loop is for the config options at api.config.general.filteredParams, because it still catches primitive parameters. I split it out to not check the config list for every param in the action, just once at the end. So once through the action's params and once through the config settings.

I could change the parameter names so it's easier to read (param params filteredParams isn't easy on the eyes).

evantahler:
Ok, fair enough! 

There are some other place where we might use dot-prop, where we are trying to find a method on the api object as defined by a nested property string.  Would you mind taking a look at those locations within this PR as well?

* https://github.com/actionhero/actionhero/blob/5a88c5912ed87702b784439c713334ce0f2e8442/initializers/actionProcessor.js#L8
* https://github.com/actionhero/actionhero/blob/dcf6dd9b529c0e844ee0bc821abe7fa75ef9aef2/initializers/redis.js#L77
* https://github.com/actionhero/actionhero/blob/746f057ca78c86603cd0610f3550bf23746b785d/initializers/i18n.js#L22

evantahler:
Perhaps we can use dot-prop to replace api.utils.stringToHash

chimmelb:
Yep, I will address these. Stay tuned!

evantahler:
Looks good to me! 
Are you ready to have this merged in?

chimmelb:
Yep, ready : )

Issue #1064, opened 2017-05-29
cluster_pidfile not removed during Gracefully stopping 


bbaga:
I was wondering why isn't the cluster_pidfile removed when the cluster shuts down gracefully.
After starting a cluster with some workers I wait for the boot sequence to finish, then press CTRL+C

```
^CGracefully stopping... (press Ctrl+C again to force)
Stopping web-api ... done  (docker stuff)
```

The pid file is still there and I get an error of course:
```
error:  Error: actionHeroCluster already running (pid xx)
```

I looked into the code and it seems there isn't even an attempt to remove the cluster's pid file and that made me wonder why is it being kept on shut down.

v16.0.0

evantahler:
Excellent idea! 
I've implemented that here https://github.com/actionhero/actionhero/pull/1065

PR #1065, opened 2017-05-29
remove cluster_pidfile when stopping; catch for removing the 0th worker

evantahler:


evantahler:
solves https://github.com/actionhero/actionhero/issues/1064

bin/methods/start/cluster.js
```diff
@@ -190,6 +190,20 @@ ActionHeroCluster.prototype.writePidFile = function (callback) {
...
+  const file = this.options.pidPath + '/' + this.options.pidfile
+
+  if (fs.existsSync(file)) {
+    const filePid = parseInt(fs.readFileSync(file))
+    if (process.pid !== filePid) {
+      return callback(new Error(`another process wrote this pid ${filePid}`))
+    }
+  }
+
+  fs.unlinkSync(file)
```
usebaz: why you use sync function and in next line call callback?
maybe better: `fs.unlink(file, callback)`?

evantahler: Really, it is just convince here.  This cluster-master process is only doing one 'thread' of work at a time (no parallel requests, blocking one operation until another completes, etc).    

Feel free to send in a PR to change it!

Issue #1074, opened 2017-06-18
Custom loader


billybonks:
I Prefer to use promises and define my tasks in a specific way that gives me pleasure, in order to do this i need to be able to load my task/action in then transform it to have the shape that actionhero framework expects it to have.

when i initially did this i tried to make it work via the config, but it seems if i put a function in the config it gets executed recursively (tried closures as well) i am currently running off a fork

https://github.com/billybonks/actionhero/commit/2093aeb6c0cfcfa4d452e84ba3b4bcdc978f5a6f#diff-af4d0f17c39abaac1efe373a4721d4ccR59

can you help me to find a way, to build this custom loader without having to fork the codebase.

this is the code i added to handle my custom task api.

my tasks look like this

```
const Task = require('../lib/task');

class SomeTask extends Task {
  async run({pullRequestOnlineId, memberId, reviewStatus}){
   }
}

module.exports = AddReview
```

if i load this task 

`SomeTask = require('./tasks/sometask')`

I can run it synchronously or asynchronously

```
sm = new SomeTask()
sm.run(params)

SomeTask.runAsync(params) //this function uses actionheros api

//  static runAsync(params){
//    debug(`running task ${this.name}`)
//    tasks.enqueue(this.name, params, 'default');
//  }
```

evantahler:
There's a lot to unpack here!
First, why would you run a task synchronously?  Are you trying to use the same methods in an action and task?  If so, put them in an initializer!   Who would `require('./tasks/sometask')` ?

However, there have been a number of times I've been tempted to re-write actionhero's API to look more like this, ES6-class-y.  Do you like this syntax better?  Right now, ActionHero looks a lot like [next](https://github.com/zeit/next.js/), in that the file-system is the convention for loading the server (tasks live in /tasks, actions in /actions, etc).  If we want to abandon that and do a more explicit "main" file which requires everything, perhaps this style makes more sense?

billybonks:
ye it is quite a bit of information, was worried about putting to much.

> Right now, ActionHero looks a lot like next, in that the file-system is the convention for loading the server (tasks live in /tasks, actions in /actions, etc). If we want to abandon that and do a more explicit "main" file which requires everything, perhaps this style makes more sense?

I am happy with the filesystem layout, but either a main file or a config file where custom loaders can be added would be great, i would go with a config for now since it seems like the easiest. 

>  ES6-class-y. Do you like this syntax better?

[...]

PR #1075, opened 2017-06-25
Custom loaders

billybonks:
Following this conversation https://github.com/actionhero/actionhero/issues/1074. @crrobinson14 @evantahler 

I made an example of what i am thinking, i think this is a good solution because it doesn't break anything for existing projects, but adds a point of customisation for those who want. it also allows me to share my customisations via ah-plugins.

`loaders/task.js` allows me to use a different syntax for my tasks, and easily change it as i require more features.

```
module.exports = function(fullFilePath) {
  const klass = require(fullFilePath);

  function makeRunner(context){
    function runner (api, params, next){
      let _task = new context.klass;
      _task.run(params).then(() =>{ next(); }).catch(next);
    }
    return runner;
  }

  let task = {
    name: klass.name,
    klass: klass,
    description:   '  ',
    queue:         'default',
    frequency:     0,
  };

  task.run = makeRunner(task);
  return [task];
};
```
`loaders/console.js` allows me to modify my repl env
```
module.exports = function(repl) {
  repl.context.test = 'asd';
};
```
i am using the default loader for actions.

### DIFF
--- a/bin/methods/console.js
+++ b/bin/methods/console.js
@@ -27,6 +27,10 @@ module.exports = {
 
         repl.context.api = api
 
+        if (api.config.loaders.console) {
+          api.config.loaders.console(repl)
+        }
+
         repl.on('exit', function () {
           next(null, true)
         })
--- a/config/api.js
+++ b/config/api.js
@@ -50,6 +50,7 @@ exports['default'] = {
         'pid': [path.join(__dirname, '/../pids')],
         'log': [path.join(__dirname, '/../log')],
         'server': [path.join(__dirname, '/../servers')],
+        'loader': [path.join(__dirname, '/../loaders')],
         'cli': [path.join(__dirname, '/../bin')],
         'initializer': [path.join(__dirname, '/../initializers')],
         'plugin': [path.join(__dirname, '/../node_modules')],
--- a/initializers/actions.js
+++ b/initializers/actions.js
@@ -67,7 +67,7 @@ module.exports = {
       let action
 
       try {
-        const collection = require(fullFilePath)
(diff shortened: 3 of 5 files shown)


evantahler:
Please check the test suite.  Some of your style doesn't pass the linter.  Also, please lock in the behavior you expect with a test.

billybonks:
will do so the overall strategy looks okay to you ?

evantahler:
I'm personally having trouble imaging a use case... but I'm sure that's because I've been stuck inside of ActionHero for too long... can you share a full example (with multiple files) in your PR's description?  The example you have now looks to me like it can be accomplished within the structure that exists already.

usebaz:
I think this is not necessary functionality. It is too specific and rarely used. And unclear

evantahler:
@billybonks Also, we would require a test added which locks-in your new functionality, which could also help explain the use-case you have.

billybonks:
Sorry for the delay on finalising this, it will be ready by Monday for
sure. Thanks for your patience :)

billybonks:
sorry again for the delay very, tied up atm. @evantahler  when I try to run the test I just get

image

evantahler:
I don't think I see what the problem is? Join us on slack if you need help!

evantahler:
I'm closing this as the next version of actionHero will require that actions and tasks are ES6 classes, which can be easily extended however you like.  Also, all run methods will be async methods, so there will be no ned for a promise-ify wrapper like your example!  

For example: https://github.com/actionhero/actionhero/pull/1098

billybonks:
awesome thanks :)

PR #1078, opened 2017-06-30
Fix  memory leak for actions with `data.toRender: false` 

witem:
For actions with `data.toRender !== true`, not called `connection.destroy()`.
This provoked memory leak.

Pruff:
* actionhero/master: http://i.imgur.com/L8wyvzH.png
* fixed version: http://i.imgur.com/9j2JGR8.png

Each picture is taken after 20k requests to actions with `data.toRender = false;`.

### DIFF
--- a/actions/randomNumber.js
+++ b/actions/randomNumber.js
@@ -14,19 +14,3 @@ exports.randomNumber = {
   }
 
 }
-
-exports.randomNumberAsync = {
-  name: 'randomNumberAsync',
-  description: 'I am an API method which will async generate a random number and custom render response',
-  outputExample: '0.123',
-
-  run: function (api, data, next) {
-    data.toRender = false
-    process.nextTick(() => {
-      data.connection.rawConnection.res.writeHead(200, { 'Content-Type': 'text/plain' })
-      data.connection.rawConnection.res.end(`${Math.random()}`)
-      next(null)
-    })
-  }
-
-}
--- a/actions/randomNumber.js
+++ b/actions/randomNumber.js
@@ -14,3 +14,19 @@ exports.randomNumber = {
   }
 
 }
+
+exports.randomNumberAsync = {
+  name: 'randomNumberAsync',
+  description: 'I am an API method which will async generate a random number and custom render response',
+  outputExample: '0.123',
+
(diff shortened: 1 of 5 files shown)


initializers/connections.js
```diff
@@ -167,6 +167,7 @@ module.exports = {
         if (typeof server.goodbye === 'function') { server.goodbye(this) }
       }
 
+      api.connections.connections[this.id] = undefined
```
evantahler: Why would you set the connection to undefined if you are going to delete it on the next line?

servers/web.js
```diff
@@ -348,63 +348,72 @@ const initialize = function (api, options, next) {
...
-    if (data.toRender === true) {
-      if (api.config.servers.web.metadataOptions.serverInformation && typeof data.response !== 'string') {
-        const stopTime = new Date().getTime()
-        data.response.serverInformation = {
-          serverName: api.config.general.serverName,
-          apiVersion: api.config.general.apiVersion,
-          requestDuration: (stopTime - data.connection.connectedAt),
-          currentTime: stopTime
-        }
+    if (data.toRender !== true) {
```
evantahler: Great find!

initializers/connections.js
witem: Set to `undefined` help GC clear data, but `delete` only delete property, not clear memory.
http://bertanguven.com/preventing-memory-leaks-in-javascript-null-vs-delete

witem:
@evantahler I add test

actions/randomNumber.js
```diff
@@ -14,3 +14,19 @@ exports.randomNumber = {
   }
 
 }
+
+exports.randomNumberAsync = {
```
evantahler: Please don't add a new action just for testing.  You can create the action within the test suite like here:  https://github.com/actionhero/actionhero/blob/master/test/servers/web.js#L416-L445

witem: done

evantahler:
Thanks! I'm merging this is now, but I'll be removing your setting the connection to `undefined`.  We *are* deleting properties of an object, so `delete` is the proper command here.

witem:
Ok, thanks!

Issue #1081, opened 2017-07-10
There is no way to access request's raw body


antonsapyanov:
I implement an integration with Stripe, and need access to request's raw body in order to verify signature (Stripe makes POST requests on specified webhook).
I have found [that](https://github.com/actionhero/actionhero/blob/v16.0.0/servers/web.js#L484) in the source code:
```
connection.rawConnection.form = new formidable.IncomingForm()

for (i in api.config.servers.web.formOptions) {
  connection.rawConnection.form[i] = api.config.servers.web.formOptions[i]
}

connection.rawConnection.form.parse(connection.rawConnection.req, (error, fields, files) => {
  if (error) {
    server.log('error processing form: ' + String(error), 'error')
    connection.error = new Error('There was an error processing this form.')
  } else {
    connection.rawConnection.params.body = fields
    connection.rawConnection.params.files = files
    fillParamsFromWebRequest(connection, files)
    fillParamsFromWebRequest(connection, fields)
  }
  if (api.config.servers.web.queryRouting !== true) { connection.params.action = null }
  api.routes.processRoute(connection, pathParts)
  callback(requestMode)
})
```
As I understood, there is no way to access raw body via `rawConnection` object. Is there any other way (maybe suggestions) how to access it?

- ActionHero Version: `16.0.0`

evantahler:
If you are talking about a JSON body payload, ie: `curl -H "Content-Type: application/json" -X POST -d '{"username":"xyz","password":"xyz"}' http://localhost:3000/api/login`, you can use `connection.rawConnection.params.body`.

evantahler:
keep in mind that `data.connection.rawConnection.params.body` differs from `data.params` in a few ways:
- no formatters, validators, defaults, etc will be applied
- no sanitization will be done on them, so be sure that you are testing that the request is coming from a trusted source (ie: only stripe's servers)

antonsapyanov:
I mean exactly **raw body** (String or Buffer), before it is parsed to an object.

antonsapyanov:
Stripe sends JSON object. There is no reason to stringify `data.connection.rawConnection.params.body`, because there is no guarantee, that result string will equal request's raw body string (which is used to make Stripe Signature header).

antonsapyanov:
Could you reopen the issue please? I think it is quite important

evantahler:
Can you provide as test CURL request with a body payload which emulates the response you are trying to access from ActionHero?

evantahler:
[...]

PR #1082, opened 2017-07-10
store connection.rawConnection.params.rawBody on web connections to keep around raw POST body

evantahler:
solves https://github.com/actionhero/actionhero/issues/1081

### DIFF
--- a/servers/web.js
+++ b/servers/web.js
@@ -507,7 +507,9 @@ const initialize = function (api, options, next) {
         }
 
         let rawBody = []
-        connection.rawConnection.req.on('data', (chunk) => { rawBody.push(chunk) })
+        if (api.config.servers.web.saveRawBody) {
+          connection.rawConnection.req.on('data', (chunk) => { rawBody.push(chunk) })
+        }
 
         connection.rawConnection.form.parse(connection.rawConnection.req, (error, fields, files) => {
           if (error) {
--- a/servers/web.js
+++ b/servers/web.js
@@ -506,9 +506,9 @@ const initialize = function (api, options, next) {
           connection.rawConnection.form[i] = api.config.servers.web.formOptions[i]
         }
 
-        let rawBody = []
+        let rawBody = Buffer.alloc(0)
         if (api.config.servers.web.saveRawBody) {
-          connection.rawConnection.req.on('data', (chunk) => { rawBody.push(chunk) })
+          connection.rawConnection.req.on('data', (chunk) => { rawBody = Buffer.concat([rawBody, chunk]) })
         }
 
         connection.rawConnection.form.parse(connection.rawConnection.req, (error, fields, files) => {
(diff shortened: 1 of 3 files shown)


servers/web.js
```diff
@@ -505,12 +505,17 @@ const initialize = function (api, options, next) {
...
+        let rawBody = []
+        connection.rawConnection.req.on('data', (chunk) => { rawBody.push(chunk) })
+
         connection.rawConnection.form.parse(connection.rawConnection.req, (error, fields, files) => {
           if (error) {
             server.log('error processing form: ' + String(error), 'error')
             connection.error = new Error('There was an error processing this form.')
           } else {
             connection.rawConnection.params.body = fields
+            connection.rawConnection.params.rawBody = Buffer.concat(rawBody).toString().substring(0, api.config.servers.web.formOptions.maxFieldsSize)
```
evantahler: normally, we would need to wait until `connection.rawConnection.req.on('end')`, but formidable does the same thing, we can wait within it's callback

evantahler:
@crrobinson14 @genexp This is likely to increase the memory footprint of every connection... good idea or bad idea?

evantahler:
I guess the thing to do here would be to make this optional!

crrobinson14:
@evantahler Aside from the length checking ability it adds, why jump through the Buffer/toString hoops to copy rawBody out? It seems like it's extra processing that may not be necessary?

Two possible options that might be an enhancement?
1. Instead of making this a global setting, make it metadata in the action. I think it's likely that only one or two API calls in most stacks will need this, so most devs probably don't need to turn it on for every call, just weird stuff like Stripe (which signs the body with a header field so you need to check the raw body).
2. Just store a ref to rawBody, but add a connection class method to obtain it, such as `.getRawBody()`. That could do the buffer/string/length limit dance only where required. This doesn't reduce the memory footprint (because a ref is saved for rawBody, of course), but defers it until other things have done. What if input validation fails, or other middleware aborts the connection? The raw body processing will have been wasted - making it a great DDOS make-work vector...

evantahler:
Excellent ideas! I'll work some of them up now.  You hit the nail on the head: a lot of this is dangerous (memory, DDOS, etc).

evantahler:
@crrobinson14 I think we would still need to store the body-buffer somewhere, and since we want it destroyed along with the connection, it seems to make sense to append it there.  Regarding the buffer-to-string, you are right! might as well let the consuming action handle that.

evantahler:
regarding action metadata, I really don't like action-metadata which is only useful for one type of connection.  We only have one of those and it tears me up inside :D

More importantly, the connection is fully parsed (and params extracted) before the action is even determined.  We can't wait until then to know if we need to keep the body around.

krlicmuhamed:
Can I read a request body Stream in some action instead of It just storing complete body data?

evantahler:
The problem is that the stream is complete by the time the action is started.  It needs to work this way because we need all FROM options/uploads, from the body stream, to be available (and added to `params`) for the action.

krlicmuhamed:
Maybe we can add another function in `exports.action` that runs before body stream starts?
That way we can have _some_ shared memory with server code. 
Something like this:
```
var rawBody;
exports.action = {
 name: 'test',
 initialize: function(shared){
  shared.events.on('rawbodydata', function(raw){
   ...
   rawBody = raw;
  });
 }
 run: function (api, data, next) {}
}
```
This is the `connection.rawConnection.params.rawBody` done in a different way though, except you gain more flexibility :)

evantahler:
I really don't like the implementation of this pull request... so send you idea as a pull request!

PR #1093, opened 2017-09-01
task.frequency can be number or function

witem:
This PR which will allow use set `tast.frequency` as function, for getting values from config.
Task frequency can be function and first argument is *api*.

Example:
```
exports.simpleTask = { 
     name: 'simpleTask', 
     description: 'I is simple task, 
     queue: 'high', 
     frequency: (api) => api.config.general.tasksFrequncy.simpleTask, 
  
     run: co.wrap(function* (api, params, next) { yield runSimpleAction (); }),
};
```

It need for different env and conditions.
Example some task I run only on stage and prod, but not on dev. And now I must in each task file paste some like this:
```
const FREQUENCY_TASK_ENABLED = process.env.NODE_ENV === 'production' || +process.env.FORCE_BACKGROUND_TASK === 1; 
const FIX_TRANSACTION_FREQUENCY = FREQUENCY_TASK_ENABLED ? 15 * 60 * 1000 : 0;

exports.simpleTask = {
    // .....
    frequency: FIX_TRANSACTION_FREQUENCY,
    // .....
};
```

- [x] run `npm test`

### DIFF
--- a/initializers/tasks.js
+++ b/initializers/tasks.js
@@ -43,12 +43,13 @@ module.exports = {
 
       jobWrapper: function (taskName) {
         const task = api.tasks.tasks[taskName]
+        const frequency = typeof task.frequency === 'function' ? task.frequency(api) : task.frequency
 
         let middleware = task.middleware || []
         let plugins = task.plugins || []
         let pluginOptions = task.pluginOptions || []
 
-        if (task.frequency > 0) {
+        if (frequency > 0) {
           if (plugins.indexOf('jobLock') < 0) { plugins.push('jobLock') }
           if (plugins.indexOf('queueLock') < 0) { plugins.push('queueLock') }
           if (plugins.indexOf('delayQueueLock') < 0) { plugins.push('delayQueueLock') }
@@ -121,7 +122,7 @@ module.exports = {
         } else if (typeof task.description !== 'string' || task.description.length < 1) {
           fail('Task ' + task.name + ' is missing \'task.description\'')
           return false
-        } else if (typeof task.frequency !== 'number') {
+        } else if (typeof task.frequency !== 'number' && typeof task.frequency !== 'function') {
           fail('Task ' + task.name + ' has no frequency')
           return false
         } else if (typeof task.queue !== 'string') {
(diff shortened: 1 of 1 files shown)


gcoonrod:
Please add tests to cover normal behavior and error cases.

witem:
@gcoonrod in AH test I have not found any test for testing periodic(with frequency) tasks. Therefore, I think testing this functional should not be included in PR. Need include mock util(like sinon) and write tests with it. But it will take a lot of time

initializers/tasks.js
```diff
@@ -43,12 +43,13 @@ module.exports = {
 
       jobWrapper: function (taskName) {
         const task = api.tasks.tasks[taskName]
+        const frequency = typeof task.frequency === 'function' ? task.frequency(api) : task.frequency
```
evantahler: The function invocation should probably be run in the same context as the task's definition, so therefore: `task.frequency.call(this, api)`

initializers/tasks.js
```diff
@@ -240,13 +241,14 @@ module.exports = {
 
       enqueueRecurrentJob: function (taskName, callback) {
         const task = this.tasks[taskName]
+        const frequency = typeof task.frequency === 'function' ? task.frequency(api) : task.frequency
```
evantahler: same comment as above, `task.frequency.call(this, api)`

initializers/tasks.js
```diff
@@ -261,7 +263,9 @@ module.exports = {
 
         Object.keys(this.tasks).forEach((taskName) => {
           const task = this.tasks[taskName]
-          if (task.frequency > 0) {
+          const frequency = typeof task.frequency === 'function' ? task.frequency(api) : task.frequency
```
evantahler: Since you are doing this check (function vs number) so often, you can probably make a helper method in this file to do it

initializers/tasks.js
evantahler: As a general rule, other methods on the api object which are referenced by task or action definitions (like middleware names) are defined as strings, rather than the method itself.  

This is important as the method might not actually be defined yet when the task is loaded into RAM.  To evaluate `"api.determineTaskFrequency"` from the string name, see how we do it here: https://github.com/actionhero/actionhero/blob/master/initializers/redis.js#L77-L89

evantahler:
Hi! I'm closing this pull request, as this functionality will be included in the next version of ActionHero.  Tasks of the future will be defined as classes, and their frequency property can either be set directly, or by a synchronous method.

This behavior was locked-in via this test https://github.com/actionhero/actionhero/pull/1088/commits/1c6825ed50517b13d6d8923552ae6362d257d652

PR #1098, opened 2017-09-18
Action classes

evantahler:
### Should Actions/Tasks/Servers/Initializers become classes?

```js
'use strict'

module.exports = class RandomNumber extends ActionHero.Action {
  constructor () {
    super()
    this.name = 'randomNumber'
    this.description = 'I am an API method which will generate a random number'
    this.outputExample = { randomNumber: 0.123 }
  }

  async run (api, data) {
    data.response.randomNumber = Math.random()
    data.response.stringRandomNumber = data.connection.localize(['Your random number is {{number}}', {number: Math.random()}])
  }
}

```

A more elaborate example with validators and inputs can be found here https://github.com/actionhero/actionhero/blob/434e9a10b0776c9b90dac1a639d7174eb9452070/actions/cacheTest.js

# Pros
* defaults set by inheritance 
* properties (like name, description, etc) can me methods or strings
* extendible actions (ie: `myAction` extends `authenticatedAction` (which extends `ActionHero.action`)
  * this allows for shared helper methods, which might be used for inputs, run methods, etc.  Shared middleware chains, etc.
* overwritable/customizable validators

# Cons
* more breaking changes
* classes spec not *quite* done in js/node yet
* probably slower to boot

### DIFF
--- a/actions/cacheTest.js
+++ b/actions/cacheTest.js
@@ -1,41 +1,56 @@
 'use strict'
+const ActionHero = require('./../index.js')
 
-exports.cacheTest = {
-  name: 'cacheTest',
-  description: 'I will test the internal cache functions of the API',
-
-  outputExample: {
-    cacheTestResults: {
-      saveResp: true,
-      sizeResp: 1,
-      loadResp: {
-        key: 'cacheTest_key',
-        value: 'value',
-        expireTimestamp: 1420953274716,
-        createdAt: 1420953269716,
-        readAt: null
-      },
-      deleteResp: true
-    }
-  },
-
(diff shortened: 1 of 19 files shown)


evantahler:
Potentially a solve for https://github.com/actionhero/actionhero/issues/1074 and an alternative to https://github.com/actionhero/actionhero/pull/1075

gcoonrod:
I like this. The empty `super()` is kind of annoying, but looking at the required parameters I don't see anything that can be readily left off.

evantahler:
Yep, we are doing this!

PR #1099, opened 2017-09-19
[Promise] Destructuring function parameters

gcoonrod:
This PR is a part of the [PR 1088](https://github.com/actionhero/actionhero/pull/1088) work being done.

### DIFF
--- a/actions/randomNumber.js
+++ b/actions/randomNumber.js
@@ -9,8 +9,8 @@ module.exports = class RandomNumber extends ActionHero.Action {
     this.outputExample = { randomNumber: 0.123 }
   }
 
-  async run (api, data) {
-    data.response.randomNumber = Math.random()
-    data.response.stringRandomNumber = data.connection.localize(['Your random number is {{number}}', {number: Math.random()}])
+  async run (_, {connection, response}) {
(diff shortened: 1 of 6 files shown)


actions/randomNumber.js
```diff
@@ -9,8 +9,8 @@ module.exports = class RandomNumber extends ActionHero.Action {
     this.outputExample = { randomNumber: 0.123 }
   }
 
-  async run (api, data) {
-    data.response.randomNumber = Math.random()
-    data.response.stringRandomNumber = data.connection.localize(['Your random number is {{number}}', {number: Math.random()}])
+  async run (_, {connection, response}) {
```
evantahler: So to me, `_` makes me think of the underscore library.  Since you have to assign it to *something*, why not just leave it as `api`... because that is the name of the object after all...

evantahler: This is what I'm doing on the CLI branch: https://github.com/actionhero/actionhero/blob/af22fcbb95fe469b4174b398f260d1f10e17859c/bin/methods/generate/action.js#L19

gcoonrod: That makes sense.

PR #1101, opened 2017-09-19
Class servers

evantahler:
```js
'use strict'

module.exports = class MyServer extends ActionHero.Server {
  constructor () {
    super()
    this.type = 'myServer'

    this.attributes = {
      canChat: false,
      logConnections: true,
      logExits: true,
      sendWelcomeMessage: false,
      verbs: []
    }

    // this.api will be set at runtime
    // this.config will be set to equal api.config.servers[this.type]
  }

  initialize () {
    this.on('connection', (conection) => {

    })

    this.on('actionComplete', (data) => {

    })
  }

  start () {

  }

  stop () {

  }

  sendMessage (connection, message, messageCount) {

  }

  sendFile (connection, error, fileStream, mime, length, lastModified) {

  }

  goodbye (connection) {

  }
}
```

Notes:
* This deletes `genericServer` as a thing as our normal class pattern here is where it moves too
* This also updates the specHelper server in the same way


initializers/servers.js
```diff
@@ -21,30 +21,35 @@ module.exports = {
...
+    for (let i in serverFolders) {
+      let p = serverFolders[i]
+      let files = api.utils.recursiveDirectoryGlob(p)
+      for (let j in files) {
+        let filename = files[j]
+        let ServerClass = require(filename)
+        let server = new ServerClass()
+        server.config = api.config.servers[server.type] // shorthand access
+        if (server.config && server.config.enabled === true) {
+          server.api = api // this is terrible, but needed pass the connection, logger, and staticFile classes on
```
evantahler: I really don't like this this.  

However, I'm not sure how to get the API object in scope for the rest of the classes' methods.  For example, the websocket server needs to refrence `api.servers.servers.web` to bind to it.  The web server wants to read the `api.config.general.serverName`, etc...

initializers/servers.js
```diff
@@ -21,30 +21,35 @@ module.exports = {
...
-          api.servers.servers[serverName] = serverObject
-          api.log(`Initialized server: ${serverName}`, 'debug')
+    for (let i in serverFolders) {
+      let p = serverFolders[i]
+      let files = api.utils.recursiveDirectoryGlob(p)
+      for (let j in files) {
+        let filename = files[j]
+        let ServerClass = require(filename)
+        let server = new ServerClass()
+        server.config = api.config.servers[server.type] // shorthand access
```
evantahler: This, on the other hand, seems fine.

evantahler:
@gcoonrod I'm going to merge this in, but I'd still love your comments!

PR #1108, opened 2017-09-22
JSDocs

evantahler:
Docs are cool!
Automatically make docs.


evantahler:
@crrobinson14 I took a crack at a more formal JSdoc setup.  Your PR was a good starting point, but for all the initializers, we don't want them exposed as classes, so I think namespaces are a better option.   

screen shot 2017-09-22 at 3 38 49 pm

However, I can't figure out how to make the method names actually say `api.cache.load` rather than just `load`.  As silly as that is, I think it is important.

evantahler:
Describing classes, even with virtual methods, makes a little more sense
screen shot 2017-09-22 at 4 40 54 pm

docs/tutorials/actions.md
```diff
@@ -0,0 +1,247 @@
+## Overview
```
evantahler: @crrobinson14 @gcoonrod  it looks like we also have the opportunity to keep our existing 'docs' pages as 'tutorials' within the docs site... which seems to make some sense

PR #1109, opened 2017-09-23
JSDoc for connections

crrobinson14:


evantahler:
I've got some questions on what should be public/private here, but I'll merge it in and we can discuss as we go.

PR #1111, opened 2017-09-26
ActionHero.api globally available via require 

evantahler:
@crrobinson14 is a genius.  

Rather than pass the `api` object around all over the application, when it is needed, we can simply `const api = require ('actionhero').api`.  There are some tricky hoops to jump though regarding the definition of our base classes, but this creates a very pleasant developer API.

This changes the method signature of `run` for actions and tasks, initializers, servers, and CLI commands. 

## Action
```js
const {Action, api} = require('actionhero')

module.exports = class MyAction extends Action {
  constructor () {
    super()
    this.name = 'myAction'
    this.description = 'myAction'
    this.outputExample = {}
  }

  async run (data) {
    api.log('yay')
    // your logic here
  }
}
```

## Task
```js
const {Task, api} = require('actionhero')

module.exports = class MyTask extends Task {
  constructor () {
    super()
    this.name = 'myTask'
    this.description = 'myTask'
    this.frequency = 0
  }

  async run (data) {
    api.log('yay')
    // your logic here
  }
}
```

## Initializer

```js
const {Initializer, api} = require('actionhero')

module.exports = class MyInitializer extends Initializer {
  constructor () {
    super()
    this.name = 'myInitializer'
    this.loadPriority = 1000
    this.startPriority = 1000
    this.stopPriority = 1000
  }

  async initialize () {
    api['myInitializer'] = {}
  }

  async start () {}
  async stop () {}
}
```

## CLI Command

```js
const {CLI, api} = require('actionhero')

module.exports = class MyCLICommand extends CLI {
  constructor () {
    super()
    this.name = 'myCLI'
    this.description = 'myCLI'
    this.inputs = {}
  }

  async run (data) {
    api.log('yay')
    return true
  }
}
```

## Server
```js
const {Server, api} = require('actionhero')

module.exports = class MyServer extends Server {
  constructor () {
    super()
    this.type = 'myServer'

    this.attributes = {
      canChat: false,
      logConnections: true,
      logExits: true,
      sendWelcomeMessage: false,
      verbs: []
    }
    // this.config will be set to equal api.config.servers[this.type]
  }

  initialize () {
    this.on('connection', (conection) => {

    })

    this.on('actionComplete', (data) => {

    })
  }

  start () {
    // this.buildConnection (data)
    // this.processAction (connection)
    // this.processFile (connection)
  }

  stop () {

  }

  sendMessage (connection, message, messageCount) {

  }

  sendFile (connection, error, fileStream, mime, length, lastModified) {

  }

  goodbye (connection) {

  }
}
```


classes/process.js
```diff
@@ -17,28 +22,25 @@ module.exports = class Process {
...
     }
 
-    this.api = {
-      running: false,
-      initialized: false,
-      shuttingDown: false,
-      projectRoot: projectRoot,
-      bootTime: null
-    }
+    api.running = false
```
gcoonrod: Is there a reason these parameters aren't namespaced? e.g. `api.actionheroInstance.running`.

index.js
```diff
@@ -24,3 +24,5 @@ const path = require('path');
 ].forEach(({klass, file}) => {
   exports[klass] = require(path.join(__dirname, 'classes', file))
 })
+
+exports.api = {}
```
gcoonrod: Each one of the classes in this file are going to require in `api` and then modify it. Why not create the `api` object at the start of the file and then run a reduce over the array of Classes and pass in `api`. That should remove the need for the weird `require` statements.

test/core/api.js
```diff
@@ -65,25 +65,25 @@ describe('Core: API', () => {
...
           }
         },
         '2': {
           name: 'versionedAction',
           description: 'I am a test',
           version: 2,
           outputExample: {},
-          run: async (api, connection) => {
-            connection.response.version = 2
+          run: async (data) => {
```
gcoonrod: Why the change from `connection` to `data`?

actions/cacheTest.js
```diff
@@ -50,7 +50,8 @@ module.exports = class CacheTest extends ActionHero.Action {
     }
   }
 
-  async run ({cache}, {params, response}) {
+  async run ({params, response}) {
+    const {cache} = ActionHero.api
```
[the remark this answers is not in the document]
evantahler: Agree! I'll go with `const {Action, api} = require('./../index.js')`

bin/actionhero
```diff
@@ -6,6 +6,7 @@ const path = require('path')
 const optimist = require('optimist')
 const spawn = require('child_process').spawn
 const ActionHero = require(path.join(__dirname, '..', 'index.js'))
+const api = ActionHero.api
```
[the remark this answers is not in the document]
evantahler: a little lower down we also do `const actionHeroProcess = new ActionHero.Process()`

bin/templates/cli.js
```diff
@@ -1,7 +1,7 @@
 'use strict'
 const ActionHero = require('actionhero')
 
-module.exports = class MyCLICommand extends ActionHero.Action {
+module.exports = class MyCLICommand extends ActionHero.CLI {
```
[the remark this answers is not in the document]
evantahler: 🤐

index.js
evantahler: The 4 top-level properties of the api object are:
```
    api.running (bool)
    api.initialized (bool)
    api.shuttingDown (bool)
    api.projectRoot (string)
    api.bootTime (Date)
```

I really do think that these belong at the top of the stack (rather than under `api.status` or something, as the do describe the API itself.

test/core/api.js
evantahler: `connection` was wrong, it's an acton so `data` is what you get.  I could have done `{response}` I guess, but I was just correcting a naming error.

classes/process.js
evantahler: The 5 top-level properties of the api object are:
```
    api.running (bool)
    api.initialized (bool)
    api.shuttingDown (bool)
    api.projectRoot (string)
    api.bootTime (Date)
```

I really do think that these belong at the top of the stack (rather than under `api.status` or something, as they do describe the API itself.

index.js
evantahler: I tried a few methods of this, but since the `package.json#main` of `actionhero` is this index file, it has to be the one exporting... something.   If I have the `Process` return the API, that means I need to instantiate it within index, and I don't want to do that either.

evantahler: If you can come up with a better pattern, by all means please submit it! I'm *happy enough* with how it works now, as only the private classes are strange.  No user-facing files have the require problem.

PR #1115, opened 2017-09-27
fix race condition in websocket server

S3bb1:
When you connect to actionhero via the client over websocket and send 2 parallel requests .... one action request and one "verb" (like 'roomAdd', 'roomLeave' or 'roomView') in the following constellation:

1. send verb roomLeave request
-> connection.messageNumber++ (now 1)
2. send action request (action takes a little longer ~500ms) 
-> connection.messageNumber++ (now 2)
3. answer received from roomLeave request (messageCount send back from actionhero: 2(!!!))
4. answer received fromaction request (messageCount send back from actionhero: 2(!!!))

When the verb is answered, no "messageCount" is given to server.sendMessage ... so the implementation takes the current connection.messageCount number to answer the request.

this pr fixes this problem with creating a temporary variable for the message count and giving it to the server.sendMessage function

### DIFF
--- a/servers/websocket.js
+++ b/servers/websocket.js
@@ -220,6 +220,7 @@ const initialize = function (api, options, next) {
     const verb = data.event
     delete data.event
     connection.messageCount++
+    const messageCount = connection.messageCount
     connection.params = {}
     if (verb === 'action') {
       for (let v in data.params) {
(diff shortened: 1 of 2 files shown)


evantahler:
This is a great find, and good solution! 

Can you please add a test that shows that this is working?  You can use the `sleepTest` action for something slow, and any verb for something fast.

evantahler:
I've added this test to the master branch which ensures the proper behavior (if I understand your test case properly) https://github.com/actionhero/actionhero/commit/0733574f6469d1b967ac6118eae8e98b8d56644e

S3bb1:
sorry for the delay, added a test for it :)

evantahler:
re-running your test suite after https://github.com/actionhero/actionhero/commit/642a862f58a93cac63fb0f124d7e91bd3cf7e89d

PR #1116, opened 2017-09-27
Migrating documentation part 1

gcoonrod:


gcoonrod:
I need to figure out what is going on with my commits here.

PR #1123, opened 2017-10-03
Update Plugins

evantahler:
### Remove `actionhero link` (and `actionhero unlink`) in favor of `config/plugins.js`
 Using linkfiles was brittle.  It didn't work with namespaced NPM packages, and struggled on windows computers.  We are returning to using a configuration file to define plugins which your application will load.  

```js
// config/plugins.js

// If you want to use plugins in your application, include them here:
return {
  'myPlugin': { path: __dirname + '/../node_modules/myPlugin' }
}

// You can also toggle on or off sections of a plugin to include (default true for all sections):
return {
  'myPlugin': {
    path: __dirname + '/../node_modules/myPlugin',
    actions: true,
    tasks: true,
    initializers: true,
    servers: true,
    public: true
  }
}
```

This also makes testing plugins much easier, as you can boot up an ActionHero server from within your plugin (if `actionhero` is a devDependancy) with the following: 

```js
const path = require('path')
const actionhero = new ActionHero.Process()
let api

describe('My Plugin', () => {
  before(async () => {
    let configChanges = {
      plugins: {
        'testPlugin': { path: path.join(__dirname, '..') }
      }
    }

    process.env.PROJECT_ROOT = path.join(__dirname, '..', 'node_modules', 'actionhero')
    api = await actionhero.start({configChanges})
  })

  after(async () => { await actionhero.stop() })

  it('does stuff', async () => {
    //...
  })
})
```

### Add `actionhero generate plugin`
A helper which you can use in an empty directory which will create a template plugin project

### Remove `api.utils.recursiveDirectoryGlob` in favor of the nom `glob` package.
We can use the standard package now that we no longer need to traverse custom ActionHero link files


bin/templates/package-plugin.json
```diff
@@ -0,0 +1,12 @@
+{
+  "author" : "YOU <YOU@example.com>",
+  "name" : "my_actionhero_plugin",
+  "description" : "my actionhero plugin",
+  "version" : "0.0.1",
```
crrobinson14: Because of a quirk in NPM's Semver I suggest starting at `0.1.0`. NPM allows updates that don't
touch the LEFT MOST non-zero number. If we are on version 0.4.3, this will allow an update to 0.4.4 but NOT 0.5.1. However, if we are on 1.4.3, and we publish 1.5.1, this WILL be brought in automatically.

The trouble with 0.0.# is that if you bump from 0.0.5 to 0.0.6 your changes will not be brought in even though instinctively this is a non-breaking change! In this case the left-most non-zero number is the patch level itself, so `npm update` won't see it.

config/plugins.js
```diff
@@ -0,0 +1,26 @@
+exports['default'] = {
+  plugins: (api) => {
+    /*
+    If you want to use plugins in your application, include them here:
+
+    return {
+      'myPlugin': { path: __dirname + '/../node_modules/myPlugin' }
```
crrobinson14: This neatly solves the scoping issue too. Nice.

docs/ActionHero.html
```diff
@@ -200,7 +200,7 @@ <h4 class="name" id="~ActionCallback"><span class="type-signature"></span>Action
     
     <dt class="tag-source">Source:</dt>
     <dd class="tag-source"><ul class="dummy"><li>
-        <a href="initializers_actions.js.html">initializers/actions.js</a>, <a href="initializers_actions.js.html#line6">line 6</a>
+        <a href="initializers_actions.js.html">initializers/actions.js</a>, <a href="initializers_actions.js.html#line8">line 8</a>
```
crrobinson14: Should we start git-ignoring the generated files and just have Travis do this generation and publication? We're going to start getting some awfully messy diffs...

initializers/tasks.js
```diff
@@ -540,10 +542,19 @@ module.exports = class Tasks extends ActionHero.Initializer {
 
     api.tasks.loadTasks = (reload) => {
       api.config.general.paths.task.forEach((p) => {
-        api.utils.recursiveDirectoryGlob(p).forEach((f) => {
+        glob.sync(path.join(p, '**', '*.js')).forEach((f) => {
           api.tasks.loadFile(f, reload)
         })
       })
+
+      for (let pluginName in api.config.plugins) {
```
crrobinson14: A little confused about a detail - you may know this instinctively because you just wrote it. What's the order of operations here if your main project overrides a file in the plugin, as it can override core today? Will that still work? It SEEMS like this is backward to support that but I wasn't sure at first glance and it's late...

bin/templates/package-plugin.json
```diff
@@ -0,0 +1,12 @@
+{
+  "author" : "YOU <YOU@example.com>",
```
gcoonrod: Could we read this value from their git config?

initializers/specHelper.js
```diff
@@ -199,6 +199,7 @@ module.exports = class SpecHelper extends ActionHero.Initializer {
         connection.actionCallbacks[(connection.messageCount)] = resolve
       })
 
+      if (!response.error) { delete response.error }
```
gcoonrod: What does this do? If there is no `error` delete it?

docs/ActionHero.html
evantahler: There's a whole other conversation we need to have about this... to Slack with it!

initializers/tasks.js
evantahler: good question! You can't override initializers in core (it throws an error), and there are no actions or tasks which core exports (just a few it dumps into a new project on `generate`).  So... there really was no thought at all given to overriding things.  Should we throw an error?  Should we allow overriding?

bin/templates/package-plugin.json
evantahler: Possible, but either you need to shell out (and that's a pain on *nix/osx vs windows), or we need to bring in another package to help... I'd rather just skip it.

initializers/specHelper.js
evantahler: There's somewhere that returning `undefined` for the error rather than `null`... I couldn't find it, but I'll look a little deeper...

initializers/tasks.js
gcoonrod: I like the idea of allowing overriding but logging a warning message.

docs/ActionHero.html
evantahler: closed by https://github.com/actionhero/actionhero/pull/1127

PR #1134, opened 2017-10-10
Update tasks.js

rakhnin:
fix task.run this context


evantahler:
Did you sort out what you needed help with?  Join us on [slack](https://slack.actionherojs.com/) if you need help!

rakhnin:
Hi @evantahler. In my vision `this` context for the task class instance itself is missing inside `run` method (One not able to call additional task methods through `this`). But it seems this hack is used in the unit tests ;(. So I need to dive deeply in this before merge request. Or I miss something else in the AH task/worker architecture?

evantahler:
conversation moved to https://github.com/actionhero/actionhero/pull/1135/

PR #1135, opened 2017-10-10
Fix `this` in tasks

rakhnin:
Update `this` context inside `run` method to the task class instance itself.

It allows to call others task instance methods inside `run`method through `this`.

```js

module.exports = class SomeTask extends ActionHero.Task {
  constructor () {
    super()
    this.name = 'someTask'
    this.description = '...'
    this.frequency = 0
    this.queue = 'default'
    this.middleware = []
  }

  async stepOne (params) {
    ...
  }

  async stepTwo (params) {
    ...
  }

  async run (params) {
    let resOne = await this.stepOne()
    let resTwo = await this.stepTwo()
    ...
  }
}

```

### DIFF
--- a/classes/task.js
+++ b/classes/task.js
@@ -39,6 +39,7 @@ module.exports = class SayHello extends Task {
   /**
    * The main "do something" method for this task.  It can be `async`.  Anything returned from this metod will be logged.
    * If error is thrown in this method, it will be logged & caught.  Using middleware, you can decide to re-run the task on failure.
+   * `this` is a Task instance itself now.
    * Instance of a node-resque worker can be accessed at this.worker within a Task's run method. This means you can inspect `this.worker.job` and set `this.worker.result` explicitly if your Task does not return a value.
    *
    * @function run
--- a/classes/task.js
+++ b/classes/task.js
@@ -37,11 +37,14 @@ module.exports = class SayHello extends Task {
   }
 
   /**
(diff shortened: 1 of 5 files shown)


evantahler:
Interesting! I just added https://github.com/actionhero/actionhero/commit/b797da4e5a9252fc3258b9a95d478ae8d2726b2c to the docs to explain what `this` in a task *is* at the moment.   Your example for why you want to change what `this` is makes perfect sense.

Thanks!

evantahler:
Can you please update the reference I mentioned above in the task description to call out explicitly that `this` is now the task itself?  This *might* be a breaking change for some folks.

rakhnin:
@evantahler done.

rakhnin:
Sorry @evantahler. Seems it is bad idea with current passing worker instance. Cause task is a singleton ;). I’ll update passing as a second argument of the ‘run’ method.

evantahler:
Great! 

Since this problem started as the `run` method of a task not properly having access to `this` (per your example), could you also please add a test that confirms this behavior?   After that, I'll merge this in!

rakhnin:
I've added the test.

evantahler:
Amazing.  Thanks!
I'll get this released today

evantahler:
If you want to talk more about ActionHero, please join the rest of us on https://slack.actionherojs.com!

rakhnin:
Thanks for the invitation ;)

PR #1136, opened 2017-10-11
specHelper redis connection

rakhnin:
This is a fix for problem running unit-tests with non-standard redis configuration. E.g.

```sh
REDIS_HOST=192.168.99.100 npm test -- test/core/tasks.js
```

It tries to reconnect with default redis parameters. See `node-resque/lib/connection`.


evantahler:
awesome, great find!

Issue #1083, opened 2017-07-13
Replace all instances of invoking actionhero with `./node_modules/.bin/actionhero/` with `npx actionhero`


evantahler:
https://medium.com/@maybekatz/introducing-npx-an-npm-package-runner-55f7d4bd282b

nicolasembleton:
Saw that yesterday and thought about exactly the same thing. That'll be a great improvement from a readability perspective.

evantahler:
I think we should at least wait until npm v5.2 is shipped with a version of node...

nicolasembleton:
Yes. Smart. It's not critical to change yet. But breaking it or others will make it a critical bug. So yeah, makes total sense. Could offer both at some point too I suppose. (curious up to which node / npm version is actionhero going to be supported?)

_Sent from my HTC U11 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

evantahler:
ActionHero will always target LTS versions of node for as long as they are supported.  So we support v4, 6 and 8 right now.  I *belive* npm attempts to always be backwards-compatible, so that shouldn't be an issue.

gcoonrod:
Wouldn't this change only impact newly created projects?

evantahler:
I think this would actually only be a change to the documentation and test suite.

evantahler:
Since ActionHero v18 requires node v8+, we can make this change now

evantahler:
solved by https://github.com/actionhero/actionhero/pull/1138

PR #1138, opened 2017-10-12
update ./node_modules/.bin/actionhero to npx actionhero

evantahler:
Now that `npm` ships with a handy command (`npx`) to run a command, we should use it in our docs.

https://medium.com/@maybekatz/introducing-npx-an-npm-package-runner-55f7d4bd282b


evantahler:
soves https://github.com/actionhero/actionhero/issues/1083

PR #1143, opened 2017-10-19
Validators returning false now emit a generic error.

gcoonrod:
Currently, any param validators that return false do not result in an error response to the action. While the documentation states that a validator should return `true` or an error, having a generic error thrown when `false` is returned results in a slightly more intuitive developer experience.


evantahler:
Love it!  
I don't know what's going wrong with the greenkeeper integration... but merging anyway!

PR #1145, opened 2017-10-26
Fixed missing error property in WebSocket message

lukaserat:


evantahler:
While this will double-send the error string down as both `status` and `error`, I think that is OK.  Can you please add a test (here: https://github.com/actionhero/actionhero/blob/master/test/servers/websocket.js) which confirms this behavior?

evantahler:
checking in again on that test!  Do you have any questions about how to do it?

classes/actionProcessor.js
```diff
@@ -201,10 +201,10 @@ module.exports = class ActionProcessor {
         let validatorResponse
         try {
           if (typeof validator === 'function') {
-            validatorResponse = await validator.call(api, params[key], this)
+            validatorResponse = await validator.call(api, params[key], this, key)
```
evantahler: Can you share why you are adding another argument to the validator method signature?  I don't think this belongs in this Pull Request.  Can you please remove it and we'll talk about it separately in a new one?

initializers/chatRoom.js
```diff
@@ -421,7 +421,7 @@ class ChatRoom extends ActionHero.Initializer {
       let toReturn = true
       if (messagePayload) { newMessagePayload = Object.assign({}, messagePayload) }
 
-      api.chatRoom.globalMiddleware.forEach(async (name) => {
+      await Promise.all(api.chatRoom.globalMiddleware.map(async (name) => {
```
evantahler: Great find! Can you please add a test that confirms that aync chat middleware will both fire?  Let me know if you need guidance on creating the test.

evantahler:
It seems you are working on things other than this patch in your branch, so I'm going to close this PR.

Issue #1149, opened 2017-11-07
HTTP Request to the web server do not work if there are no parameters set and just the action.


Martinlck:
# Name of Issue
Empty parameters request sent to web do not work properly
Description of Issue
Server does not reply, throws error when it should not.

- ActionHero Version: `latest`
- Node.js Version: `9.0.0` also 8
- Operating System:  `Ubuntu 14`

# Steps to reproduce your error
Install action hero, generate empty project.
Start action hero.
Go to: "localhost:8080/showDocumentation" 
Nothing happens, and you see some rejection error in some promise and a message error on the console
if you try localhost:8080/showDocumentation?asd   it works fine

Hello guys, something broke on the new action hero version, this was working fine on 17, I migrated to 18 and I'm having issues with web requests that don't need any parameters.

Thanks.

evantahler:
What setting do you have for `api.config.servers.web.rootEndpointType`?  The default is `file`, which means that files in your `public` directory are those that will shown for root paths.  To hit one of your actions (assuming that `api.config.servers.web.simpleRouting = true`) you would need the API path, ie `localhost:8080/api/showDocumentation`.  Adding a query parameter to the request has no effect. 

Without the `api`, you should actually see the error as defined by the file server `That file is not found`, set by `api.config.errors.fileNotFound` and `locales/en.json` .

![screen shot 2017-11-09 at 11 01 08 am](https://user-images.githubusercontent.com/303226/32624164-4cb148e6-c53d-11e7-8678-eee935ec67ab.png)

In checking older versions of ActionHero (v17) I can confirm that the behavior is the same, and adding a query parameter to the request has no effect.

Martinlck:
Hello @evantahler I'm sorry my mistake, the url is actually localhost:8080/api/showDocumentation 
And it does not work it throws the error.

Martinlck:
Ok after researching about the issue, found out this: https://github.com/nodejs/node/issues/13404
So is non related to the AH version, rather AH not working on GET requests with action and not QS on latest node versions.
I will submit a PR linking to this issue.

evantahler:
Oh! great find!  Thank you for taking the time to submit a fix!

Martinlck:
Thank you for building such an amazing API server.

evantahler:
will be solved by https://github.com/actionhero/actionhero/issues/1149#issuecomment-343530353

PR #1152, opened 2017-11-10
Fixed slicing the search of parsed url when only receiving an action on QS.

Martinlck:
Hello guys,
I tried to follow the guideline, please let me know if you want me to change something.

The issue I'm trying to fix on this commit is: https://github.com/actionhero/actionhero/issues/1149
which is related to this node fix: nodejs/node#13404

I wrote a test which is pretty similar to the rest, since the tests were not passing anyway on node 9. 

After this fix all tests pass again.

Thanks

### DIFF
--- a/servers/web.js
+++ b/servers/web.js
@@ -474,7 +474,12 @@ module.exports = class WebServer extends ActionHero.Server {
     // API
     if (requestMode === 'api') {
       if (connection.rawConnection.method === 'TRACE') { requestMode = 'trace' }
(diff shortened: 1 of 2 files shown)


evantahler:
Oh! This is great! I didn't realize that we had problems on node v9.  I've used your fix here (https://github.com/actionhero/actionhero/pull/1153) and ensured that we test v8 and v9 of Node.JS on CI.

Thank you!

Issue #1155, opened 2017-11-17
Compound Action with Shared Inputs code does not work as it should


SpeedyC86:
# Tutorial code not working correctly
Compound Action with Shared Inputs code does not work as it should.
Not returning true in the validator function will create a magical issue that will not run the "run" function code. Please correct the tutorial and make the "should return true if validation passed" bold :) It was hard debugging and trying to find out why the code was not working.

- ActionHero Version: `latest`
- Node.js Version: `8.9.1`
- Operating System: `(Ubuntu 16)`

# Steps to reproduce your error
The code in the run function will not be run.
```
// Compound Action with Shared Inputs//
const {Action, api} = require('actionhero')

class ValidatedAction extends Action {
  constructor () {
    super()
    this.inputs = {
      email: {
        required: true,
        validator: this.emailValidator
      },
      password: {
        required: true,
        validator: this.passwordValidator
      }
    }
  }

  emailValidator (param) {
    if (email.indexOf('@') < 0) {
      throw new Error('that is not a valid email address')
    }
  }

  passwordValidator (param) {
    if (param.length < 4) {
      throw new Error('password should be at least 3 letters long')
    }
  }
}

// the actions
exports.UserAdd = class UserAdd extends ValidatedAction {
  constructor () {
    super()
    this.name = 'userAdd'
    this.description = 'I add a user'
  }

  run (data) {
    // your code here
  }
}

exports.UserDelete = class UserDelete extends ValidatedAction {
  constructor () {
    super()
    this.name = 'userDelete'
    this.description = 'I delete a user'
  }

  run (data) {
    // your code here
  }
}
```
---

evantahler:
Thank you for this!  

I think the best behavior would actually be to allow the behavior you thought was possible in the first place.  Only if a validator returns `false` or throws an error should the param count as invalid.  

PR to update this here https://github.com/actionhero/actionhero/pull/1158

evantahler:
closing in favor of https://github.com/actionhero/actionhero/pull/1158

PR #1158, opened 2017-11-21
action validators which return nothing are valid

evantahler:
Changes the behavior of action's param validators to only mark a param as invalid if they return `false` or throw an error.  Previously, validators which did not return (or returned null) would be marked as failing. 

Solves https://github.com/actionhero/actionhero/issues/1155.

### DIFF
--- a/classes/actionProcessor.js
+++ b/classes/actionProcessor.js
@@ -206,6 +206,11 @@ module.exports = class ActionProcessor {
             const method = this.prepareStringMethod(validator)
             validatorResponse = await method.call(api, params[key], this)
           }
+
+          // validator function returned nothing; assume param is OK
+          if (validatorResponse === null || validatorResponse === undefined) { return }
+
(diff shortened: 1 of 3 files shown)


evantahler:
@gcoonrod do you see this as a breaking change, or a clarification of what was previously undefined behavior?

gcoonrod:
This is a change of current behavior. If existing users relied on this behavior it would be a breaking change.

evantahler:
Dang... so this would be a major SEMVER change in your opinion?

gcoonrod:
Do actions failing after an upgrade with no other code change amount to a breaking change? The documentation does state that returning nothing is a valid response from a validator. So this change brings the behavior in line with the documentation. I could go either way honestly. Though I'm leaning more towards a minor version change.

evantahler:
v18.1.0 it is!

PR #1159, opened 2017-12-01
Issue with unclosed file descriptors

derstrassi:
Found an issue in the action hero web server. If the web server sends a status code 304 response, then the fileStream handle is not closed.
200s are not affected, because their streams get automatically closed by pipe().

### DIFF
--- a/classes/server.js
+++ b/classes/server.js
@@ -142,7 +142,7 @@ module.exports = class MyServer extends Server {
    * @memberof ActionHero.Server
    * @param  {Object}  connection The connection in question.
    * @param  {Error}   error If there was any errror finding or reading this file.
-   * @param  {Object}  fileStream A stream from the file reader which can be piped to the connection.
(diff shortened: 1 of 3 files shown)


evantahler:
This is a great find! 

I initially wanted to see if there was a way to generically add this kind of check into `initializers/staticFile`, but I don't think that's a possibility.  Only the connection's server will know if the file's stream was actually used by anyone.  Right now, only the web server has this problem (there's no cache mechanism (etag, etc) for the websocket or socket servers).

So it looks like it really is the responsibility of the server to ensure that the `filestream` really is either `pipe`'d or closed.  Can you please add a note about this to the docs (https://github.com/actionhero/actionhero/blob/master/classes/server.js#L140)?

derstrassi:
Made a small note about closing the fileStream if it is not piped.

evantahler:
Thanks!

PR #1168, opened 2017-12-30
Support node.js v9.x with ActionHero v17

evantahler:
Bring in the changes from https://github.com/actionhero/actionhero/pull/1153 to node v17


evantahler:
Released in v17.1.2

PR #1177, opened 2018-02-11
WIP: ActionHero + Serverless

evantahler:
This PR is a WIP demonstrating some of the ideas about running ActionHero in a "serverless" (AWS Lambda, etc) environment. 

1. `npm install --global serverless`
2. checkout this branch
3. disable the servers we don't need (web, websocket) by changing `enabled = false` in their config files. 

Now, you can run your actions!

* Run an action: `serverless invoke local -f randomNumber`
* Run an action with params: `serverless invoke local -f cacheTest -l --data '{"key": "key", "value": "value"}'`
* Get a file (base-64 encoded) `serverless invoke local -f file -l --data '{"file": "simple.html"}'`

TODO:
- This runs locally on node.js v8+... but AWS only supports v6.  You *can* ship custom binaries with Lambda functions... we should figure out how.
- Is there a way to auto-generate `serverless.yml` from `api.actions.actions`?
- Tests... somehow.

### DIFF
--- a/handler.js
+++ b/handler.js
@@ -1,0 +1,128 @@
+// const { Process, Server, api }  = require('actionhero')
+const { Process, Server, api } = require('./index.js')
+
+const actionhero = new Process()
+const ReturnMetadata = true
+
+class LambdaServer extends Server {
+  constructor () {
+    super()
+    this.type = 'lambda'
+    this.attributes = {
+      canChat: false,
+      logConnections: false,
+      logExits: false,
+      sendWelcomeMessage: false,
+      verbs: []
+    }
+  }
+
+  start (api) {
+    this.on('connection', (connection) => { this.handleConnection(connection) })
+    this.on('actionComplete', (data) => { this.actionComplete(data) })
+  }
+
+  stop () {}
+
+  goodbye () {}
+
+  async handleConnection (connection) {
+    if (typeof connection.rawConnection.event === 'object') {
+      connection.params = connection.rawConnection.event
+    }
+
+    const parts = connection.rawConnection.context.functionName.split('-')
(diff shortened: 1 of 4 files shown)


evantahler:
This is generally on hold until AWS Lambda supports node v8+

handler.js
```diff
@@ -0,0 +1,24 @@
...
+const awaitRunning = async () => {
+  await sleep()
+  if (api.running !== true) {
+    return awaitRunning()
+  }
+}
+
+(async () => {
+  module.exports.run = async (event, context, callback) => {
+    await awaitRunning()
```
gcoonrod: Is this a check to wait for the server to finish starting up?

servers/lambda.js
```diff
@@ -0,0 +1,131 @@
...
+
+  goodbye () {}
+
+  async handleConnection (connection) {
+    if (typeof connection.rawConnection.event === 'object') {
+      connection.params = connection.rawConnection.event
+    }
+
+    const parts = connection.rawConnection.context.functionName.split('-')
+    const actionName = parts[(parts.length - 1)]
```
gcoonrod: These two lines seem brittle...

handler.js
evantahler: Yeah... the weird thing here is that the "entrypoint" here is this exported function, so this method is the first thing that is called, *then* you need to boot actionhero :/

servers/lambda.js
evantahler: very! good call.

chimmelb:
Woot! 
https://aws.amazon.com/blogs/compute/node-js-8-10-runtime-now-available-in-aws-lambda/

PR #1180, opened 2018-02-15
generic error decoration hook for action errors, CLI docs cleanup

chimmelb:
As discussed [in Slack channel](https://actionherojs.slack.com/archives/C04EVSUSD/p1518021134000659), Actions that throw an error get a generic hander in `api.config.errors.genericError`. Here the `data.response` could be decorated however needed, like added a `requestId`, or whatever. Default is pure passthrough.

Added the same for a plain `status` that is not an error, just for consistency. Default is passthrough. 

Few tweaks to CLI documentation.

### DIFF
--- a/config/errors.js
+++ b/config/errors.js
@@ -83,6 +83,16 @@ exports['default'] = {
         return api.i18n.localize(['actionhero.errors.dataLengthTooLarge', {maxLength: maxLength, receivedLength: receivedLength}])
       },
 
+      // Decorate your response based on Error here
+      async genericError (data, error) {
+        return error
+      },
+
+      // Decorate your response based on status value
+      async genericStatus (data, status) {
+        return status
+      },
+
       // ///////////////
       // FILE SERVER //
       // ///////////////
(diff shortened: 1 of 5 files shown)


config/errors.js
```diff
@@ -83,6 +83,16 @@ exports['default'] = {
         return api.i18n.localize(['actionhero.errors.dataLengthTooLarge', {maxLength: maxLength, receivedLength: receivedLength}])
       },
 
+      // Decorate your response based on Error here
+      async genericError (data, error) {
+        return error
+      },
+
+      // Decorate your response based on status value
+      async genericStatus (data, status) {
```
evantahler: I don't think I understand what this does.  You also haven't included a test for it.

config/errors.js
```diff
@@ -83,6 +83,16 @@ exports['default'] = {
         return api.i18n.localize(['actionhero.errors.dataLengthTooLarge', {maxLength: maxLength, receivedLength: receivedLength}])
       },
 
+      // Decorate your response based on Error here
+      async genericError (data, error) {
```
evantahler: 👍 for non-breaking changes!

config/errors.js
evantahler: /status/ is an internal actionProcessor thing normally

evantahler:
Thanks!

PR #1184, opened 2018-02-24
Custom mime handling

chimmelb:
From [Slack chat today](https://actionherojs.slack.com/archives/C04EVSUSD/p1519406808000282)

I have an odd use case, where legacy app is setting `Content-Type: text/xml` (or `text/plain`). This is a POST which is handled [by this code](https://github.com/actionhero/actionhero/blob/master/servers/web.js#L472-L500), ultimately being processed by `node-formidable`, which [results in an error](https://github.com/felixge/node-formidable/blob/c7e47cd640026d12b64b9270ecb60f6c2585c337/lib/incoming_form.js#L289). By the look of current AH code, any non-GET, non-HEAD request that sets a Content-Type (including`application/json`) gets passed to `node-formidable`.

Regarding `api.config.servers.web.saveRawBody`, if fomidable errors, the saved rawBody is a 0-byte Buffer. (Formidable might pause the `req` processing)

This PR gives the option to handle req processing in a custom manner, if one so chooses. If the API receives any mime types defined in `config/servers/web.js:tailoredMimeTypes[]`, a handler is run that allows a hook for custom processing. (In my case a "do nothing" actually works, because I just want `connection.rawConnection.params.rawBody` to be saved/filled.)  A user could put an entire `text/plain` body into `connection.params.text`, or parse XML, or whatever they want that formidable won't do (since it's for forms : )

I will add tests, but wanted feedback on this approach. Current tests do pass currently (`392 passing (2m)`), so this is a non-breaking change as-is, and would give the option to custom process web requests from non-formidable MIME types.


evantahler:
I'm adding a few other folks as reviewers to this PR, as I'm conflicted about what to do: 
- From the slack chat, I *think* that this PR is in response to a technically invalid request... the request *shouldn't* be XML, just the payload... I think.  
- That said, allowing folks to handle real-world situations, even if they are out-of-spec is useful.
  - If this is out-of-spec, rather than making an option, can we create an overridable situation instead?
- Or is this really a bug?

Either way, I think seeing the test first would make the failing scenario much clearer, then we can discuss the implementation.  TDD FTW!

config/servers/web.js
```diff
@@ -6,6 +6,7 @@ exports['default'] = {
   servers: {
     web: (api) => {
       return {
+        '_toExpand': false,
```
gcoonrod: What is this doing?

servers/web.js
```diff
@@ -469,31 +469,38 @@ module.exports = class WebServer extends ActionHero.Server {
...
-        let {fields, files} = await new Promise((resolve) => {
-          connection.rawConnection.form.parse(connection.rawConnection.req, (error, fields, files) => {
-            if (error) {
-              this.log('error processing form: ' + String(error), 'error')
-              connection.error = new Error('There was an error processing this form.')
-            }
-            resolve({fields, files})
+        // Check for custom handling of mime type
+        let mimeType = connection.rawConnection.req.headers['content-type'] || connection.rawConnection.req.headers['Content-Type']
+        if (this.config.tailoredMimeTypes.indexOf(mimeType) > -1) {
```
gcoonrod: I would prefer to see Array.prototype.includes() used here as it is less confusing to future readers.

config/servers/web.js
chimmelb: It allows the functions to get merged as config settings are compiled. Used in `config/errors.js` and `config/redis.js` too (for `api.utils.hashMerge`)

chimmelb:
Thanks for thinking about this. I made an update (`indexOf()` -> `includes()`) and added some tests. I think the issue at hand is a bit of bullets 1, 2 and 3.

Mostly this is bullet 2, where I have an outside system that I want to talk to AH. That system sends text as the request body and sets the request's `Content-Type` to "text/xml". If it was my own API this wouldn't be an issue, as JSON works great!

That's where it is a bit of bullet 1 and 3. AH doesn't like POSTing data on the Web server that isn't handled with formidable. The check [happens here](https://github.com/actionhero/actionhero/blob/master/servers/web.js#L464-L471), which makes a new IncomingForm and parses. I see the error on L485 about a bad form when I used my curl requests OR my legacy app. If I'm reading it correctly, AH would never handle a POST that specifies a `Content-Type` header (that formidable doesn't handle natively).

I'm open to an override option rather than this config-style approach. Hopefully those tests explain a bit as well. 

I wouldn't expect this to be widely used, but for those rare cases it's at least a hook to handle it, rather than no options like I'm seeing now. (If there's another option, I'm all ears! : )

config/servers/web.js
gcoonrod: I'm not a fan of an option name that needs to be in single quotes because it isn't a valid javascript key. Why does this need to be the option key?

evantahler: It doesn't, but that's how it has been for over a year.  A change now would be a breaking change, unrelated to this PR.  

It's a signal to `api.utils.hashMerge` & `api.utils. isPlainObject` to *not* attempt to expand a function and hydrate it

chimmelb:
I updated this to not require a change in `config/servers/web.js`, to avoid breaking changes. If the array of custom mime types is not found, or does not contain the request's mime type, process as normal using formidable.

Regarding the `'_toExpand'` thing, it would not need single quotes. Underscores are valid JS properties, I guess the quoting thing just isn't caught by standardJS rules?

evantahler:
Ok, here's a crazy idea:  Rather than provide custom parser options as config files... perhaps all we need to do is not error if the content isn't actually JSON.  

I could imagine 1 new config option `api.config.servers.web.allowInvalidJson`, and if it is set to `true`, we'll catch the formidable error set set `connection.rawConnection.params.rawBody`.   From there, you can parse the body however you want in a middleware or action.

evantahler:
I added some tests around this to https://github.com/actionhero/actionhero/pull/1196.

> Regarding api.config.servers.web.saveRawBody, if fomidable errors, the saved rawBody is a 0-byte Buffer. (Formidable might pause the req processing)

I think the second new test in #1196 shows that this isn't true?

chimmelb:
I'm OK with just storing the rawBody if Formidable errors, though would like to try a change to how `saveRawBody` is done. Could I make an update to how we `saveRawBody` to be more like this? (in a new PR?)

```
let rawBody = await new Promise((resolve, reject) => {
            let fullBody = Buffer.alloc(0)
            if (!connection.rawConnection || !connection.rawConnection.req) {
              return resolve(fullBody)
            }
            connection.rawConnection.req
              .on('error', (err) => { reject(err) })
              .on('aborted', () => { reject(new Error('Request aborted')) })
              .on('data', (chunk) => { fullBody = Buffer.concat([fullBody, chunk]) })
              .on('end', () => { resolve(fullBody) })
          })
```

I'll rearrange the promise so it doesn't block/await before formidable. I could not get it to happen in the unit tests, but have had times when formidable errors (like we're planning) but the rawBody was not filled, even though the `saveRawBody` was set. I don't know if formidable paused/ended the stream or what. I don't think this is the culprit of my manual testing, but it's there:
https://github.com/felixge/node-formidable/blob/master/lib/incoming_form.js#L54


Another needed test is when the body is not JSON (like you have) **and** the `Content-Type` header is `text/plain` or something that formidable doesn't know how to handle.  It is kind of a different result, and more the use-case for my legacy system or AWS SNS that are setting the content type. See this one: https://github.com/actionhero/actionhero/pull/1184/files#diff-f5572dd2458d60fadaabbbffe99c04d8R423

evantahler:
If you can't reproduce a behavior in a test, please do not change anything around.  

I think my main confusion about your use-case is around:

> Another needed test is when the body is not JSON (like you have) and the Content-Type header is text/plain or something that formidable doesn't know how to handle. It is kind of a different result, and more the use-case for my legacy system or AWS SNS that are setting the content type.

check this new test: https://github.com/actionhero/actionhero/commit/a100399ab0aba38da9e13c4fe0611fb707af5511.  This is testing an a plain text payload, and ensuring that `rawBody` is set.  What else is missing?

chimmelb:
I'll have to find a test, and I like not having to write custom methods in a config.  

There is behavior that I see on local server use that prevents me from processing `rawConnection.params.rawBody`. The two places that I _think_ need addressed are 

1. how we capture `rawBody` (if 'data' event does not come, it is possible for formidable to error (unknown header), and middleware to run on an empty data)
1. If formidable errors, `connection.error` is set, but should not be if we capture `saveRawBody`? 

Maybe I can make a new PR if I can show a test that helps explain this more. Stay tuned.

chimmelb:
Took a while to get there, thanks for all the discussion. 

I think this PR could be closed to be replaced by #1198

evantahler:
closing in favor of #1198

PR #1189, opened 2018-03-02
api.routes.registerRoute for "all" methods

alimvedats:
api.routes.registerRoute method could check for “all”


evantahler:
This will require a test to ensure the functionality added.

Issue #1192, opened 2018-03-06
Action names - with dots "." Not supported


fguijarro:
# Action names - with dots "." Not supported

- ActionHero Version: v18.1.2
- Node.js Version: 8

Our team just realized that the latest action hero version does NOT support dots "." in the action names (e.g. users.add). This is a big problem for us, as that's the way we have everything. Anyone knows if this is something that will change in a future release? 

Any help/information will be very much appreciated.

evantahler:
I *think* a more specific way to describe this error would be to say that the Web Server's simple routing can't handle actions with "." in their names.

For example, the template action `random.number`

```js
const { Action } = require('./../index.js')

module.exports = class RandomNumber extends Action {
  constructor () {
    super()
    this.name = 'random.number'
    this.description = 'I am an API method which will generate a random number'
    this.outputExample = { randomNumber: 0.123 }
  }

  async run ({connection, response}) {
    response.randomNumber = Math.random()
    response.stringRandomNumber = connection.localize(['Your random number is {{number}}', {number: Math.random()}])
  }
}
```

and the routes:
```js
exports['default'] = {
  routes: (api) => {
    return {
      get: [
        { path: '/random/number', action: 'random.number' }
      ]
    }
  }
}
```

certainly work with `curl "localhost:8080/api/random/number"`

Can you confirm that this is specifically what you are seeing?

evantahler:
The source of the error when using simple routing can be found here: https://github.com/actionhero/actionhero/blob/master/servers/web.js#L355-L363

In essence, requesting a GET to a url like `/api/random.number` looks like you are requesting a file with the extension "number".  In this case, ActionHero will attempt to return a response not with the MIME of `application/json`, but whatever the mime of your file extension is.  This allows flexible APIs for things like returning text, xml, images, etc.

This behavior is actually configurable with the [`matchExtensionMimeType`](https://github.com/actionhero/actionhero/blob/master/classes/action.js#L17) property of actions.  It defaults to true, you probably want to switch that to false.

Either way, there is a bug here which fails if we can't determine a MIME/type.  The proper behavior should be to stick with `application/json` in that case.

PR #1193, opened 2018-03-06
 do not crash if we cannot find the mime/type for this action and matchExtensionMimeType is enabled

evantahler:


evantahler:
Resolves https://github.com/actionhero/actionhero/issues/1192

PR #1195, opened 2018-03-09
Adds gitignore to the copied files list

NealHumphrey:
Fixes issue https://github.com/actionhero/actionhero/issues/1188

Copies the root `.gitignore` file into generated projects. 

@evantahler I added the lines as you pointed out - pretty straightforward, thanks for the direction. I added the test and all tests pass; but, I couldn't figure out how to run the generate command locally. I saw the `sample.sh` file in the project root, but running that did not create the `.gitignore` file in the tmp directory - but it looked like maybe that would run from the npm installed version of actionhero instead of my clone of the repository? Any hints on how to do a quick manual check of this functionality using my local repo?

### DIFF
--- a/bin/methods/generate.js
+++ b/bin/methods/generate.js
@@ -109,7 +109,7 @@ module.exports = class Generate extends ActionHero.CLI {
       '/README.md': 'readmeMd',
       '/test/example.js': 'exampleTest',
       '/locales/en.json': 'enLocale',
-      '/.gitignore':'gitignore'
+      '/.gitignore': 'gitignore'
     }
 
     for (let file in newFileMap) {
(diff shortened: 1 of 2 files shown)


evantahler:
Awesome! Thanks!

evantahler:
Awsome, thanks!

Yeah, to test this locally, you'll need to go though a few hoops.  We can't use NPM link, as it doesn't link binary files... so we can do it manually:

```bash
# make a test folder
mkdir ~/Desktop/test && cd ~/Desktop/test
npm init #(just agree to all the defaults)
mkdir node_modules
# symlink in local actionhero to node_modules
ln -s ~/PROJECTS/actionhero/actionhero node_modules/actionhero
# symlink the actionhero binary (normally happens on npm install)
mkdir node_modules/.bin && cd node_modules/.bin
ln -s ../actionhero/bin/actionhero actionhero
cd ../..
# try to generate a new project
./node_modules/.bin/actionhero generate
```

PR #1197, opened 2018-03-16
Jest Testing

evantahler:
`jest` >= `mocha`.  We use it now!
* interactive testing
* test on file change
* more popular
* parallel testing

This PR Changes:
* all tests moved to `__tests__`, per Jest convention
* update all test files to work with jest and jasmine-style matchers
* removes the dev dependencies: `mocha`, `chai`, `dirty-chai`, and `cross-env`
* adds the dev dependencies: `jest`, `jest-environment-webdriver`
* updates the allowed global variables (for standard.js testing) as provided by jest
* updates `actionhero generate` to use jasmine in new projects
* adds `chromedriver` as a dev dependency; test cookies and sessions in a real browser with selenium
  * a guide on using chromedriver can be found [here](https://blog.evantahler.com/testing-javascript-applications-with-selenium-async-await-and-jest-7580ed074f2b)
* Updated various configs to use `process.env.JEST_WORKER_ID` in tests to allow parallel testing without cross-pollution of redis, servers, etc.

Also, during this test upgrade, we noticed that the way that we were sharing session between WS and WEB connections (headers) was unlikely to work with some primus/websocket transports.  We've changed to using query params which is what Priums recommends. 
* Added a new `ActionHeroWebSocket` option (generated automatically) `cookieKey`.  This *must* now be added to your config files to share sessions between WEB and WEBSOCKET connections in the same browser

Note: Due to the way that Jest overrides `require` to allow for mocking modules, we cannot test developerMode any more.  This felt like an OK tradeoff for now... 

--- 
TODO: 
* [x] Update in-project docs
* [ ] Update ActionHero example project
* [ ] Update angular CSRF project
* [ ] Update www.actionhero.com


config/servers/socket.js
```diff
@@ -29,7 +29,7 @@ exports.test = {
     socket: (api) => {
       return {
         enabled: true,
-        port: 1001 + (process.pid % 64535),
+        port: 15000 + parseInt(process.env.JEST_WORKER_ID || 0),
```
evantahler: note a test config change to allow for parallelism

config/redis.js
```diff
@@ -1,6 +1,6 @@
 let host = process.env.REDIS_HOST || '127.0.0.1'
 let port = process.env.REDIS_PORT || 6379
-let db = process.env.REDIS_DB || 0
+let db = parseInt(process.env.REDIS_DB || process.env.JEST_WORKER_ID || 0)
```
evantahler: note a test config change to allow for parallelism (multiple redis databases)

evantahler: note a test config change to allow for parallelism (multiple redis DBs)

config/servers/socket.js
evantahler: note a test config change to allow for parallelism (multiple ports)

config/servers/web.js
```diff
@@ -109,7 +109,7 @@ exports.test = {
     web: (api) => {
       return {
         secure: false,
-        port: process.env.PORT || 1000 + (process.pid % 64535),
+        port: 18080 + parseInt(process.env.JEST_WORKER_ID || 0),
```
evantahler: note a test config change to allow for parallelism (multiple ports)

initializers/cache.js
```diff
@@ -1,7 +1,6 @@
 'use strict'
 
 const fs = require('fs')
-const {promisify} = require('util')
```
evantahler: `promisify` had some weird problems with jest-mocking... so we removed it.

package.json
```diff
@@ -71,18 +68,23 @@
...
   },
+  "jest": {
+    "testPathIgnorePatterns": [
+      "<rootDir>/__tests__/testPlugin"
+    ]
+  },
   "scripts": {
     "postinstall": "echo 'To generate a new actionhero project, run \"npx actionhero generate\"'",
     "help": "node ./bin/actionhero help",
-    "test": "cross-env NODE_ENV=test mocha",
```
evantahler: jest manages NODE_ENV=test automatically, so we can drop `cross-env`!

servers/websocket.js
```diff
@@ -175,8 +172,8 @@ module.exports = class WebSocketServer extends ActionHero.Server {
   }
 
   handleConnection (rawConnection) {
-    const parsedCookies = this.fingerprinter.parseCookies(rawConnection)
-    const fingerprint = parsedCookies[api.config.servers.web.fingerprintOptions.cookieKey]
+    // console.log({headers: rawConnection.query})
+    const fingerprint = rawConnection.query[api.config.servers.web.fingerprintOptions.cookieKey]
```
evantahler: note the new way to check for passed fingerprints

__tests__/utils/utils.js
```diff
@@ -206,50 +202,50 @@ describe('Utils', () => {
...
       api.config.general.filteredParams.push('p1', 'p2', 'o2')
       let filteredParams = api.utils.filterObjectForLogging(inputs)
-      expect(filteredParams.p1).to.equal('[FILTERED]')
-      expect(filteredParams.p2).to.equal('[FILTERED]')
-      expect(filteredParams.o2).to.equal('[FILTERED]') // entire object filtered
-      expect(filteredParams.o1).to.deep.equal(testInput.o1) // unchanged
+      expect(filteredParams.p1).toEqual('[FILTERED]')
+      expect(filteredParams.p2).toEqual('[FILTERED]')
+      expect(filteredParams.o2).toEqual('[FILTERED]') // entire object filtered
+      expect(filteredParams.o1).toEqual(testInput.o1) // unchanged
```
gcoonrod: `toEqual()` is equivalent to `to.deep.equal()` as well as `to.equal()`?

bin/methods/generate.js
```diff
@@ -76,7 +76,7 @@ module.exports = class Generate extends ActionHero.CLI {
       '/public/css',
       '/public/logo',
       '/tasks',
-      '/test'
+      '/__tests__'
```
gcoonrod: Why the folder rename?

bin/methods/start/lib/actionheroCluster.js
```diff
@@ -82,7 +81,7 @@ module.exports = class ActionHeroCluster {
     if (stats.isDirectory() || stats.isSymbolicLink()) {
       return true
     } else {
-      await promisify(fs.mkdir)(p)
+      fs.mkdirSync(p)
```
gcoonrod: Why change this to a blocking method?

bin/templates/package.json
```diff
@@ -12,29 +12,26 @@
     "ioredis" : "latest"
   },
   "devDependencies" : {
-    "cross-env" : "latest",
-    "chai" : "latest",
-    "dirty-chai" : "latest",
-    "mocha" : "latest",
```
gcoonrod: I like that we can reduce our dependencies!

client/ActionheroWebsocketClient.js
```diff
@@ -37,17 +37,18 @@ ActionheroWebsocketClient.prototype.defaults = function () {
 ActionheroWebsocketClient.prototype.connect = function (callback) {
   var self = this
   self.messageCount = 0
+  var urlWithSession = self.options.url + '?' + self.options.cookieKey + '=' + self.getCookie(self.options.cookieKey)
```
gcoonrod: What if any of these parts are null or undefined?

initializers/cache.js
```diff
@@ -322,6 +321,10 @@ class Cache extends ActionHero.Initializer {
      * @private
      */
     api.cache.checkLock = async (key, retry, startTime) => {
+      const sleep = async function (time) {
```
gcoonrod: You have this defined elsewhere why not import it here?

__tests__/utils/utils.js
evantahler: Yep! https://facebook.github.io/jest/docs/en/expect.html#toequalvalue

bin/methods/generate.js
evantahler: That's JEST convention... I figure we should use the tool's defaults.

bin/methods/start/lib/actionheroCluster.js
evantahler: It's simpler to read.  Since this is s boot method, there's no real gain with the /await/.

bin/templates/package.json
evantahler: I KNOW!

initializers/cache.js
evantahler: Good call! I'll make that a util.

client/ActionheroWebsocketClient.js
evantahler: good catch! added some checks

PR #1198, opened 2018-03-17
Capture full rawBody even if formidable fails

chimmelb:
(if this gets accepted, definitely need to `squash` all the commits!)

Quick test can be run with `npm run grepTest "mime"`  (I use that locally in my `package.json`, and can remove if you don't want it in master)

Added processing that awaits for 'end' event from rawBody.

Tests try to simulate what a (slow/actual) network will do for data. If formidable fails, the connection processing finishes **before** the (full) request body is processed. In that case, `rawConnection.params.rawBody` is not available (empty) in middleware or actions. Tests also try to show that normal JSON/formidable processing still works and errors do not hang.

Thoughts on this? There are some comments I will remove, but wanted to check on the approach first.

I used this middleware when processing with my legacy stuff, simply looking for action input `mimeText`

```
var RawMimeMiddleware = {
      name: 'rawMime',
      global: false,
      priority: 1010,
      preProcessor: function (data) {
        if (data.connection.rawConnection && data.connection.rawConnection.params &&
            data.connection.rawConnection.params.rawBody) {
          data.params.mimeText = data.connection.rawConnection.params.rawBody.toString()
        }
      }
    }

    api.actions.addMiddleware(RawMimeMiddleware)
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,6 @@
     "postinstall": "echo 'To generate a new actionhero project, run \"npx actionhero generate\"'",
     "help": "node ./bin/actionhero help",
     "test": "cross-env NODE_ENV=test mocha",
-    "grepTest": "cross-env NODE_ENV=test mocha --grep $1",
     "pretest": "standard",
     "start": "node ./bin/actionhero",
     "docs": "jsdoc -c jsdoc-conf.json"
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "postinstall": "echo 'To generate a new actionhero project, run \"npx actionhero generate\"'",
     "help": "node ./bin/actionhero help",
     "test": "cross-env NODE_ENV=test mocha",
+    "grepTest": "cross-env NODE_ENV=test mocha --grep $1",
     "pretest": "standard",
     "start": "node ./bin/actionhero",
     "docs": "jsdoc -c jsdoc-conf.json"
--- a/servers/web.js
+++ b/servers/web.js
@@ -479,26 +479,12 @@ module.exports = class WebServer extends ActionHero.Server {
         if (this.config.saveRawBody) {
           connection.rawConnection.req.on('data', (chunk) => { rawBody = Buffer.concat([rawBody, chunk]) })
         }
-        // let rawBody = Promise.resolve(Buffer.alloc(0))
-        // if (this.config.saveRawBody) {
-        //   rawBody = new Promise((resolve, reject) => {
-        //     let fullBody = Buffer.alloc(0)
-        //     if (!connection.rawConnection || !connection.rawConnection.req) {
-        //       return resolve(fullBody)
-        //     }
-        //     connection.rawConnection.req
-        //       .on('error', (err) => { reject(err) })
-        //       .on('aborted', () => { reject(new Error('Request aborted while saveRawBody')) })
-        //       .on('data', (chunk) => { fullBody = Buffer.concat([fullBody, chunk]) })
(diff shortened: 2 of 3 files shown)


servers/web.js
```diff
@@ -475,9 +475,18 @@ module.exports = class WebServer extends ActionHero.Server {
...
-          connection.rawConnection.req.on('data', (chunk) => { rawBody = Buffer.concat([rawBody, chunk]) })
+          rawBody = new Promise((resolve, reject) => {
+            let fullBody = Buffer.alloc(0)
+            connection.rawConnection.req
+              .on('data', (chunk) => { fullBody = Buffer.concat([fullBody, chunk]) })
+              .on('end', () => { resolve(fullBody) })
+          })
         }
 
         let {fields, files} = await new Promise((resolve) => {
```
chimmelb: Does setting `connection.error` do anything? Does not seem to affect anything (I thought it did before), but now I don't even know where it goes....

package.json
```diff
@@ -83,6 +83,7 @@
     "postinstall": "echo 'To generate a new actionhero project, run \"npx actionhero generate\"'",
     "help": "node ./bin/actionhero help",
     "test": "cross-env NODE_ENV=test mocha",
+    "grepTest": "cross-env NODE_ENV=test mocha --grep $1",
```
evantahler: please remove this

servers/web.js
evantahler: I don't think so.  where is that code?  I don't see it referenced here.

servers/web.js
```diff
@@ -475,9 +475,18 @@ module.exports = class WebServer extends ActionHero.Server {
           connection.rawConnection.form[i] = this.config.formOptions[i]
         }
 
-        let rawBody = Buffer.alloc(0)
+        // let rawBody = Buffer.alloc(0)
+        // if (this.config.saveRawBody) {
+        //   connection.rawConnection.req.on('data', (chunk) => { rawBody = Buffer.concat([rawBody, chunk]) })
+        // }
```
evantahler: please remove these comments

test/servers/web.js
```diff
@@ -4,12 +4,13 @@ const chai = require('chai')
...
 const expect = chai.expect
 chai.use(dirtyChai)
-
+const {PassThrough} = require('stream')
 const request = require('request-promise-native')
 const fs = require('fs')
 const os = require('os')
 const path = require('path')
 const {promisify} = require('util')
+// const {PassThrough} = require('stream')
```
evantahler: please remove this comment

servers/web.js
chimmelb: Sorry, the PR review didn't let me comment outside where there were changes. It's this line: https://github.com/actionhero/actionhero/blob/master/servers/web.js#L487  Nothing needs to be changed, but I just wanted to highlight it if there was dead code (or it wasn't updated with new ways of doing things...)

chimmelb:
PR comments addressed.

Thanks again for sticking with this one. My "problems" in that previous PR (#1184) and even the approach with the "use configs to custom define handling" were all addressed here. Formidable would fail and the action processing would continue, sometimes with and sometimes without the data of my legacy system.

Hopefully this is a good change for others too, like those Amazon SNS messages and their `text/plain` payloads : )

servers/web.js
evantahler: Great question... I can't quite tell either.  We'll look into it here: https://github.com/actionhero/actionhero/issues/1199

Issue #1207, opened 2018-04-25
Chat middleware functions not bubbling up errors appropriately.


Martinlck:
# Chat middleware functions are not bubbling up errors.

Description of Issue:

runMiddleware function is not handling promises appropriately and it has some nasty hardcoded variable called toReturn. But this variable does not bubble up the errors from the promises because the original promise from the runMiddleware call resolves with whatever value was assigned to newMessagePayload.

So if join function on chatMiddleware throws error, connection that triggered that error, does not receive the error and joins the room anyway.

- ActionHero Version: latest
- Node.js Version: 9
- Operating System: All

# Steps to reproduce your error
- Make a new actionhero project with `npx actionhero generate`
- create a chat middleware and on the join async function throw error.
- connect through telnet and join a room
- observe

I'm attaching the documentation part where it describes the expected behavior ( that we are not achieving at the moment )

---
![screen shot 2561-04-25 at 5 54 08 pm](https://user-images.githubusercontent.com/1448175/39241778-37601c80-48b2-11e8-8c5a-86042970b7c5.png)

Martinlck:
Current implementation.

`    /**
     * @private
     */
    api.chatRoom.runMiddleware = async (connection, room, direction, messagePayload) => {
      let newMessagePayload
      let toReturn = true
      if (messagePayload) { newMessagePayload = Object.assign({}, messagePayload) }
      api.chatRoom.globalMiddleware.forEach(async (name) => {
        const m = api.chatRoom.middleware[name]
        try {
          if (typeof m[direction] === 'function') {
            if (messagePayload) {
              let data = await m[direction](connection, room, newMessagePayload)
              if (data) { newMessagePayload = data }
            } else {
              await m[direction](connection, room)
            }
          }
        } catch (error) {         
          toReturn = error
        }
      })
      if (toReturn !== true) {        
        return toReturn
      }
      return newMessagePayload
    }`

Martinlck:
Current implementation.

`
    /**
     * @private
     */
    api.chatRoom.runMiddleware = async (connection, room, direction, messagePayload) => {
      let newMessagePayload
      let toReturn = true
      if (messagePayload) { newMessagePayload = Object.assign({}, messagePayload) }
      api.chatRoom.globalMiddleware.forEach(async (name) => {
        const m = api.chatRoom.middleware[name]
        try {
          if (typeof m[direction] === 'function') {
            if (messagePayload) {
[...]

PR #1210, opened 2018-04-26
Fixed issue #1207. Chat middleware not bubbling up appropriately.

Martinlck:
Pointing to issue : #1207 
.forEach built in function does not work properly with async functions. Replaced as per suggestions to use for loop in series.

Updated tests that did not had the async directive on the middleware functions ( causing the tests to pass erroneously ) After putting async directive to wrap up the call in promise, the issue was reproduced.

Then after adding the suggested code with @chimmelb the test pass again and the issue is corrected.

### DIFF
--- a/__tests__/core/chatRoom.js
+++ b/__tests__/core/chatRoom.js
@@ -284,7 +284,7 @@ describe('Core', () => {
         test('(say) can modify message payloads', async () => {
           api.chatRoom.addMiddleware({
             name: 'chat middleware',
-            say: (connection, room, messagePayload) => {
+            say: async (connection, room, messagePayload) => {
               if (messagePayload.from !== 0) { messagePayload.message = 'something else' }
               return messagePayload
             }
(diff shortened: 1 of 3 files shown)


Martinlck:
Based on our Slack discussion, I reverted the test changes, and I replicated them all with async middleware functions, proving that we still have backwards compatibility and that the issue is fixed.

chimmelb:
I like these changes. More tests and less code : )

[This call](https://github.com/actionhero/actionhero/blob/3b677cae75e642d62e84084019137df24eecfa26/initializers/chatRoom.js#L144) to `api.chatRoom.runMiddleware` just awaits the error, whereas [this call](https://github.com/actionhero/actionhero/blob/3b677cae75e642d62e84084019137df24eecfa26/initializers/chatRoom.js#L194) could probably remove the `instanceof Error` check, just for clarity. I would guess the tests work without that check, since that `toReturn` stuff was removed and Errors are thrown, not returned.  

I'd say ready go after that, nice work 👍

Martinlck:
Took care of all the cases where the instanceof was being used.
It should be good to go. Thank you for the help. 
Let me know if further issues or comments.

PR #1213, opened 2018-04-26
forEach() handling of async changed to for..of

chimmelb:
Changes to allow `await` to occur in intended/existing `async` functions (not anonymous functions inside `.forEach`)


evantahler:
Thank you!

PR #1214, opened 2018-04-26
Fix routes override on multiple config folders

### LINKED ISSUE #1202: Actions override when working with multiple config/routes.js files
# Actions override when working with multiple `config/routes.js` files
When you run ActionHero with multiple `config` folders, all actions in a duplicated verb (`GET`, `POST`, `PUT`, `DELETE`) among `route.js` files get overridden.

- ActionHero Version: `18.1.3`
- Node.js Version: `8.11.1`
- Operating System: `Windows 10`

# Steps to reproduce your error
- Make a new actionhero project with `npx actionhero generate`
- Create a second `config` folder and define another `route.js` file in there
- On both `route.js` files define a route on the same verb (e.g. `POST`)
- Run ActionHero with: `ACTIONHERO_CONFIG=PATH1,PATH2 npm start`
- Try to consume both of your defined routes (one should return 404)

walbertoibarra:
Fixes #1202


config/routes.js
```diff
@@ -27,6 +27,10 @@ exports['default'] = {
 
       ---------------------- */
 
+      get: [
+        { path: '/api-status', action: 'status' }
+      ]
+
```
evantahler: Please don't check in routes file contents

walbertoibarra: Should I create  2 new config directories instead?

evantahler:
This is great, thank you! This will be part of the v19 release https://github.com/actionhero/actionhero/releases/tag/v19.0.0

Issue #1226, opened 2018-06-05
Test logger should log to a file by default


evantahler:
# Test logger should log to a file by default
When running tests, we hide all log messages from the console.  This is nice, as the test runner (jest) is the output we want to see.  That said, when something goes wrong, it is hard to debug it.  We should keep the file logger in tests, but write to a test-environment file.

- ActionHero Version: `19.x.x`
- Node.js Version: `*`
- Operating System: `*`

# Steps to reproduce your error
Run your test suite... don't see logs.

PR #1227, opened 2018-06-07
Tests now log to a file by default

etorrejon:
Logs are limited in size and number.

Resolves https://github.com/actionhero/actionhero/issues/1226


evantahler:
Thank you for this work!  I like your choice to limit the test log file size.
You have a lot of work here to generate the log directory... but that's not needed! Winston will do that for us.

config/logger.js
```diff
@@ -3,6 +3,16 @@
 const fs = require('fs')
 const cluster = require('cluster')
 
+function createDirectory (path) {
```
evantahler: Winston will do this (and exit if it can't) for us

config/logger.js
```diff
@@ -19,24 +29,21 @@ exports['default'] = {
     }
 
     // file logger
-    logger.transports.push(function (api, winston) {
-      if (api.config.general.paths.log.length === 1) {
+    const hasLogDirectoryConfigured = (api.config.general.paths.log.length === 1)
```
evantahler: I like wrapping the path check into a method!

config/logger.js
```diff
@@ -53,8 +60,27 @@ exports['default'] = {
...
+
+    if (hasLogDirectoryConfigured) {
+      logger.transports.push(function (api, winston) {
+        const logDirectory = api.config.general.paths.log[0]
+
+        createDirectory(logDirectory)
+
+        return new (winston.transports.File)({
+          filename: api.config.general.paths.log[0] + '/' + api.pids.title + '.log',
+          maxsize: 20480,
```
evantahler: Great idea!

config/logger.js
etorrejon: thanks! :)

config/logger.js
etorrejon: :)

etorrejon:
Thanks for the review, @evantahler !! I'll update

evantahler:
Thanks! Let us know if you need any help.

evantahler:
Fascinating! It looks like Winston will create any log files it needs... but not the directory if it doesn't exist.  

So, do you A) create the directory within the initializer or B) create the directory as part of the test setup (https://github.com/actionhero/actionhero/blob/master/.circleci/config.yml#L13). ?

etorrejon:
Creating in the initializer feels like a good place for it - I think that way it'll create the dir even if you're not running tests...

If that works for you, I'll add a commit that does that.

etorrejon:
Actually I guess it'd be in `config/logger.js` since that's where the transport is configured

etorrejon:
... unless you're thinking it'd be cleaner to leave that code out of config by keeping it in the build script

etorrejon:
ok sounds good! Sorry for the delay - will have a PR up soon.

evantahler:
Thansk!

Issue #1224, opened 2018-06-01
Client and Server notion of MessageCount (the "id" of the message) should be synchronized


evantahler:
# Client and Server notion of MessageCount (the "id" of the message) should be synchronized
WebSocket and Socket clients keep their own notion of MessageCount to order responses from the server.  The server does the same thing... these IDs should be in sync! 

- ActionHero Version: `19.x.x`
- Node.js Version: `all`
- Operating System: `all`

evantahler:
In a nutshell:
* client generates messageId however it wants
* client sends messageId with each request (could be a number, string... doesn't matter)
* server stores this onto each running request thread/actionProcessor
  * server checks for uniqueness of messageId
* server responds to client with messageId

evantahler:
closed by https://github.com/actionhero/actionhero/pull/1231

PR #1231, opened 2018-06-28
client and serve match messageId

evantahler:
solves https://github.com/actionhero/actionhero/issues/1224

Clients can provide `messageId` in various ways, and then have that value returned to them with their response when calling an action.  This allows better linking of responses to requests for persistent connections.

Adds a new restricted param, `messageId` is added.
This is a **breaking change** because `response.messageCount` is renamed to `response.messageId` for all servers.

Returning connectionId or not to clients is configured by `api.config.servers.web.metadataOptions.requesterInformation` on the web, and will always be returned by the `socket` and `websocket` server types

## websocket:
*warning* If you set `messageId` in one action/request, you will need to set it for EVERY request.  We will not generated it again for you.  This is the default behavior of the web client.

```js
client = new ActionheroWebsocketClient()
client.connect(function(error, details){
  client.action('randomNumber', {messageId: 'abc123'}, (data) => {
    console.log(data.randomNumber)
    console.log(data.messageId)
  }) 
}
```

## socket:
*warning* If you set `messageId` in one action/request, you will need to set it for EVERY request.  We will not generated it again for you.
```bash
> telnet localhost 5000
{"welcome":"Hello! Welcome to the actionhero api","context":"api"}

# sticky params
paramAdd messageId=abc123
{"status":"OK","context":"response","messageId":"1298be12-4873-48ee-9be9-ee407d11040b"}
randomNumber
{"randomNumber":0.9554813433986074,"stringRandomNumber":"Your random number is 0.9554813433986074","context":"response","messageId":"abc123"}
paramDelete messageId
{"status":"OK","context":"response","messageId":"abc123"}

#json params
{"action":"sleepTest","params":{"sleepDuration":5000,"messageId":"slow"}}
{"action":"sleepTest","params":{"sleepDuration":500,"messageId":"fast"}}

{"sleepStarted":1530159935733,"sleepEnded":1530159936234,"sleepDelta":501,"sleepDuration":500,"context":"response","messageId":"fast"}
{"sleepStarted":1530159935424,"sleepEnded":1530159940429,"sleepDelta":5005,"sleepDuration":5000,"context":"response","messageId":"slow"}
```

## web:
*note* a web request's `id` is now equal to `${fingerprint}-${messageId}`.  We never had a name for the second partial before!

```bash
> curl  'http://localhost:8080/api/randomNumber?messageId=abc123'

{
  "randomNumber": 0.3805910916166355,
  "stringRandomNumber": "Your random number is 0.3805910916166355",
  "serverInformation": {
    "serverName": "actionhero",
    "apiVersion": "19.0.2",
    "requestDuration": 0,
    "currentTime": 1530160113376
  },
  "requesterInformation": {
    "id": "65aed86f9a990d3bfb524640225804668dcf21c5-cbf95e66-20e7-443f-8ef6-02ae4db729dd",
    "fingerprint": "65aed86f9a990d3bfb524640225804668dcf21c5",
    "messageId": "cbf95e66-20e7-443f-8ef6-02ae4db729dd",
    "remoteIP": "127.0.0.1",
    "receivedParams": {
      "messageId": "abc123",
      "action": "randomNumber"
    }
  }
}
```

## specHelper:

```js
let response = await api.specHelper.runAction('randomNumber', {messageId: 'aaa'})
expect(response.messageId).toEqual('aaa')
```


gcoonrod:
If I understand this correctly `messageCount` was an incrementing value that kept a running tally of the number of requests the server had received? Why replace it with `messageId` and have a breaking change? Why not keep `messageCount` and add `messageId`?

gcoonrod:
@evantahler update?

evantahler:
@gred a few reasons for the breaking change:
1. the term `messageCount` assumes that the IDs need to be incremental integers, which isn't the case.  the new `messageId` is more clear that "whatever" works
2. "count" also seems to imply that there will be some validation on the string.  There won't be

I guess we could have kept both around, but they really will be doing the same thing, so I thought it would be best to be clear.  For 90% of folks, this upgrade won't be breaking, and for those that it is... I want to call it out very clearly with a major version cahnge

PR #1236, opened 2018-07-19
simplify web request-type determination and allow for pathed urlPathForActions and urlPathForFiles

evantahler:
Allows `api.config.servers.web.urlPathForFiles` and `api.config.servers.web.urlPathForFiles` to be set to sub-routes, like `/namespace/api` and `/namespace/public`


__tests__/utils/utils.js
```diff
@@ -181,6 +181,32 @@ describe('Utils', () => {
     })
   })
 
+  describe('utils.arrayStartignMatch', () => {
```
gcoonrod: Typo: `utils.arrayStartingMatch`

initializers/utils.js
```diff
@@ -393,6 +393,22 @@ let responses = await api.utils.asyncWaterfall(jobs)
       return filteredParams
     }
 
+    /**
+    Compare the first n elements of an array with another, logner array
+
+    @param {Array} a The first array, the shorter one
+    @param {Array} b The second array, the longer one
+    */
+    api.utils.arrayStartignMatch = (a, b) => {
```
gcoonrod: This assumes that the arrays are sorted. Maybe add that assumption to the jsdoc.

__tests__/utils/utils.js
evantahler: 👍

initializers/utils.js
evantahler: Interesting... are arrays not assumed to be sorted?

evantahler: eh, either way, comment added in fe52475dfa04653f5cfcfacf69549f48c875bef7 !

evantahler:
@gcoonrod thanks for the feedback!  Waiting on your review

gcoonrod:
@evantahler Looks like we're getting some intermittent test failures.

evantahler:
re-tested and green

PR #1239, opened 2018-07-27
winston v3

evantahler:
Upgrade to winston v3

* moves configuration options for logger into `config/log.js` for explicit customization
  * multiple loggers can be defined with unique levels and message formatting options
* ensures `api.config.general.paths.log` paths are created automatically (in initializer rather than config)
* reduces the logic in config file, and makes test env look a lot more like default env
* allows for multiple storage of logs in all directories defined by `api.config.general.paths.log`

This is a breaking change which is mitigated by copying in `config/log.js` from master or a new actionhero project.

### DIFF
--- a/bin/methods/start/lib/actionheroCluster.js
+++ b/bin/methods/start/lib/actionheroCluster.js
@@ -22,27 +22,32 @@ module.exports = class ActionHeroCluster {
       }
     }
 
-    let transports = []
-    transports.push(
-      new (winston.transports.File)({
-        filename: this.options.logPath + '/' + this.options.logFile
-      })
-    )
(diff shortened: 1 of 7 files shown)


evantahler:
The tests are failing because Jest uses JSDOM, which *just* published v11.12.  In Jsdom@11.12, they added localstorage (and session storage, etc)... which really hates being run in the way that Jest uses it.

We could lock our version of Jsdom to v11.11, or see how jest will handle this

Jsdom commit: https://github.com/jsdom/jsdom/commit/3afbc0feb7b2b4ac4c84530ea6af90e1f79a88ee

evantahler:
related to https://github.com/facebook/jest/issues/6766

PR #1241, opened 2018-08-08
added maxFileSize to formOptions

DaniilBabanin:
allow uploads >200mb after node-formidable v1.2.1 config update:
https://github.com/felixge/node-formidable/pull/464#issuecomment-374898543

### DIFF
--- a/config/servers/web.js
+++ b/config/servers/web.js
@@ -66,7 +66,8 @@ exports['default'] = {
         formOptions: {
           uploadDir: os.tmpdir(),
           keepExtensions: false,
-          maxFieldsSize: 1024 * 1024 * 100
+          maxFieldsSize: 1024 * 1024 * 100,
+          maxFileSize: 1024 * 1024 * 1000
         },
         // Should we pad JSON responses with whitespace to make them more human-readable?
         // set to null to disable


evantahler:
Thanks for this update!  However, I think a default setting of 1GB, when formidable itself suggests a default of 2MB, is a bit extreme, no?

DaniilBabanin:
The default form.maxFileSize has been set to [200MB](https://github.com/felixge/node-formidable/compare/v1.2.0...master#diff-38e857ca3aca5cc608230f4e8d0b3ea5R28). We have been using actionhero for various file sizes well above 200MB.

the form.maxFieldsSize is[ 20MB]( https://github.com/felixge/node-formidable/compare/v1.2.0...master#diff-38e857ca3aca5cc608230f4e8d0b3ea5R27):

gcoonrod:
I think it might be better to match the formidable default of 200MB and add a configuration override that allows setting the value higher.

evantahler:
I agree with @gcoonrod.  Match formidable's defaults, and we'll merge it in!

evantahler:
Thanks! We merged this in and are updating the default options to match formidable's defaults

Issue #1244, opened 2018-08-21
api.redis.doCluster() race conditions


vladbalmos:
# Race conditions can occur when calling api.redis.doCluster() and waiting for the response
Description of Issue

- ActionHero Version: `19.0.2`
- Node.js Version: `8.11.3`
- Operating System: `Ubuntu 16.04`

Due to waiting for the `api.redis.publish(payload)` in the `api.redis.doCluster` method, the response callback is sometimes added after the subscriber has processed the message, making the call to fail with "Error: RPC timeout"

Excerpt from `api.redis.doCluster()` (comments added for explanation):

      api.log('Publishing payload: ' + JSON.stringify(payload));
      await api.redis.publish(payload)
      // At this point, in my application, the subscriber sometimes has already processed the payload 
      // and returned a response, but the whole call will fail because the rpcCallback bellow, hasn't been
      // set yet.

      if (waitForResponse) {
        let response = await new Promise((resolve, reject) => {
          // This timer is sometimes set after the subscriber has processed and returned a response
          // making the call to eventually fail
          let timer = setTimeout(() => { 
            //api.log("RPC Timeout for " + JSON.stringify(payload));
            reject(new Error('RPC Timeout'));
          }, api.config.general.rpcTimeout)
          api.redis.rpcCallbacks[requestId] = {timer, resolve, reject}
          // api.log("Added response error timeout for payload " + JSON.stringify(payload));
        })

        return response
      }

A possible fix would be to remove the `await` keyword from the redis.publish() call, this seemed to fix the issue, but I'm not sure about other complications that it might lead.

I cannot provide tests, since I am on a tight schedule to upgrade to the async version of the framework.
    
# Steps to reproduce your error
The issue doesn't come up consistently. I was able to detect the problem by adding logging statements in `initializers/redis.js` and calling doCluster() multiple times in succession.

I hope I provided you with enough information.

evantahler:
Neat!

While I don't think we should skip the `await` (this would cause an uncahught rejection if we can't write to redis), we can ensure that the callback/promise is ready before wrting. 

Solved in https://github.com/actionhero/actionhero/pull/1246 maybe

evantahler:
Closing in favor of https://github.com/actionhero/actionhero/pull/1246

PR #1246, opened 2018-08-23
pre-init doCluster callback before writing to redis

evantahler:
solves https://github.com/actionhero/actionhero/issues/1244

Redis is so fast! Node is so Fast!  It is possible that while `await`ing the write response from redis, that another node got the job and did it already.  If we are waiting for a response, we would miss it!

This PR ensures that there is a response handler/promise before we write to redis, and that we aren't pre-`await`ing it.


gcoonrod:
Is there any way to test this?

PR #1247, opened 2018-08-23
Another fix for do cluster #1246

vladbalmos:
I tried your fix but my tests suite fails due to the same issue. I propose moving the publish call in the return promise. That seems to fix it.

### DIFF
--- a/initializers/redis.js
+++ b/initializers/redis.js
@@ -111,7 +111,6 @@ await api.redis.publish(payload)
       // we need to be sure that we build the response-handling promise before sending the request to Redis
       // it is possible for another node to get and work the request before we resolve our write
       // see https://github.com/actionhero/actionhero/issues/1244 for more information
-      let responsePromise
       if (waitForResponse) {
         return new Promise(async (resolve, reject) => {
           let timer = setTimeout(() => reject(new Error('RPC Timeout')), api.config.general.rpcTimeout)
--- a/initializers/redis.js
+++ b/initializers/redis.js
@@ -118,9 +118,9 @@ await api.redis.publish(payload)
           try {
             await api.redis.publish(payload)
           } catch (e) {
-            clearTimeout(timer);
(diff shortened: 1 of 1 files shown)


initializers/redis.js
```diff
@@ -111,19 +111,21 @@ await api.redis.publish(payload)
...
       // it is possible for another node to get and work the request before we resolve our write
       // see https://github.com/actionhero/actionhero/issues/1244 for more information
-      let responsePromise
       if (waitForResponse) {
-        responsePromise = new Promise((resolve, reject) => {
+        return new Promise(async (resolve, reject) => {
           let timer = setTimeout(() => reject(new Error('RPC Timeout')), api.config.general.rpcTimeout)
           api.redis.rpcCallbacks[messageId] = {timer, resolve, reject}
+          try {
+            await api.redis.publish(payload)
```
evantahler: This will double publish the paylod, no?

initializers/redis.js
```diff
@@ -111,19 +111,21 @@ await api.redis.publish(payload)
...
+          } catch (e) {
+            clearTimeout(timer)
+            delete api.redis.rpcCallbacks[messageId]
+            throw e
+          }
         })
       }
 
       await api.redis.publish(payload)
```
evantahler: Should you then remove this publish, or have an `if` around it?

initializers/redis.js
vladbalmos: If i'm not overlooking something, it shouldn't. If waitForResponse is true, then doCluster returns the promise with the  `try { wait publish } catch (...)` and skips the next publish call.

initializers/redis.js
vladbalmos: If `waitForResponse` is true, then this `publish` won't be executed

evantahler: Ah! I missed the return!

PR #1248, opened 2018-08-23
Wait for chat room members to be removed before destroying room

vladbalmos:
`api.chatRoom.destroy` doesn't wait for each each member to be
removed before destroying the room. This leads to chat room middleware
being run after deletion and bugs can occur if the `leave` middleware
needs to interact with the room before exiting.

Added `await` keyword to the `removeMember`.


evantahler:
awesome! Thank you!

PR #1250, opened 2018-08-29
update standard.js to v12

evantahler:


evantahler:
No, I didn't type all this `standard --fix` did it for me!

PR #1261, opened 2018-10-10
V18

cecofreeman:
Hello
In classes/connections.js, method destroy() , lines 152-155:

    if (this.canChat === true) { 
      for (let i in this.rooms) {
        let room = this.rooms[i]
        await api.chatRoom.removeMember(this.id, room)
      }
    }
If we have two rooms in this.rooms and we await for api.chatRoom.removeMember(this.id, room) the connection is removed from this.rooms and at the next iteration i becomes 1, but the length of this.rooms is 1!

I suggest to modify the code like this:

   if (this.canChat === true) { 
     let promises = []
     for (let i in this.rooms) {
       let room = this.rooms[i]
       promises.push(api.chatRoom.removeMember(this.id, room))
     }
     await Promise.all(promises)
   }
Thank you

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@ public/javascript/ActionheroWebsocketClient.js
 public/javascript/ActionheroWebsocketClient.min.js
 
 #JSDoc
-docs
+/docs
 gh-pages-branch
 
 # Node / NPM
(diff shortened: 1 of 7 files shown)


evantahler:
This pull request contains old cold that has since been removed from the project.  Please create a new pull request with these changes and re-submit

evantahler:
oh, sorry... you are actually merging work from v19 to v18 that is a documentation change... but not the part you wanted

PR #1262, opened 2018-10-11
Connections are not removed from the chat rooms

cecofreeman:
If we have two rooms in `this.rooms` and we await for `api.chatRoom.removeMember(this.id, room)` the connection is removed from this.rooms and at the next iteration` i `becomes 1, but the length of this.rooms is 1!


evantahler:
Your code does not run, you are missing a ")".  Please test your code before submitting

cecofreeman:
Hello,
Really sorry! Will not repeat...

evantahler:
closed via https://github.com/actionhero/actionhero/commit/d1960eec19f6a2129e94a8c93fca281c53337651

PR #1265, opened 2018-11-01
Fix loading a locked item

cecofreeman:
An item is loaded from the cache and is eventually old copy, because first is loaded and then is checked for locking. In such a case the locker worker may apply changes to the item during the lock, but we don't receive a fresh copy for the item. This fix is checking for lock first and then tries to load the item.

### DIFF
--- a/initializers/cache.js
+++ b/initializers/cache.js
@@ -151,10 +151,10 @@ class Cache extends ActionHero.Initializer {
      */
     api.cache.load = async (key, options) => {
       if (!options) { options = {} }
-            
+
       let lockOk = await api.cache.checkLock(key, options.retry)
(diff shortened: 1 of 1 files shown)


evantahler:
This looks like a good fix! Can you please provide a test that proves the behavior (and so it won't break in this way again?)

Thanks!

cecofreeman:
Hello,
Thank you for the comment!
Will try to provide the test within couple of days.

Regards

evantahler:
Checking in again! I've updated your branch with the latest changes from master

evantahler:
closing in favor of https://github.com/actionhero/actionhero/pull/1410

Issue #1283, opened 2019-01-11
Empty objects in Config File Merge are assigned, not merged


shookie:
# Empty objects in Config File Merge are assigned, not merged

When multiple config files are merged later empty objects under key are not merged but overwrite/clear earlier non-empty objects under the same key.

It may be possible that this behavior is wanted, but it is highly inconvenient and can be a nasty surprise.

- ActionHero Version: `19.1.1`
- Node.js Version: `8.x.x`
- Operating System: `all`

# Steps to reproduce your error

file config/A.js
```javascript
exports.default = {
  key:  { 
      A: 1
  }
}
```
and they key in file B is an empty object
file config/B.js
```javascript
exports.default = {
  key:  { 
  }
}
```
then key will be empty once all configurations are parsed and loaded, because B.js is alphabetically loaded after A.js 

- to show this do a `console.dir(api.config.key)` in any start method

The problem is in the api.utils.hashMerge function where for both arguments the

`        if (api.utils.isPlainObject(a[i]) && Object.keys(a[i]).length > 0) `

condition falls through if the object is empty, and you end up assigning in the last branch.

`    c[i] = b[i] `

the empty object case should be its own branch and just do nothing.

evantahler:
Can you share more about the use case in your configs in which the difference between `undefined` and `null` are useful?  I'm facinated!

Either way, the requested change here seems ok! Do you need help creating the pull request and tests for this?

evantahler:
sorry! I misread your bug... yes, I agree that full empty objects should be "passthroughs" on configs.  Do you need help creating the pull request and tests for this?

shookie:
Ok, maybe we misunderstand each other, and I should explain myself better. But maybe it really is the wanted behavior. But I have already heard from others that ran into issues from this merging behavior.

The use case isn't really about `null` or `undefined`, for both of which it is completely fine if later values overwrite earlier values. The use case is about empty objects being treated differently than non-empty-objects on merge.

The whole purpose of having the configuration distributed over many files is to allow grouping of configuration values that logically belong together into files, but still have the whole merged configuration available at run time.

So, you define a configuration structure that all the different config files must adhere to, because the code expects the values at the locations it expects them.

Maybe a more instructive example would help, the test case I defined is just the minimal test case, but it doesn't demonstrate why it is needed.


```
// config/departmentA.js
exports.default = {
  user:  { 
     john: { email: "john@company.com", rights: 0x1a77 },
[...]

PR #1284, opened 2019-01-14
empty config object merge fix

shookie:
Fix to #1283.


evantahler:
The test suite is failing with your changes.  I've added an additional test that explains why in 940c2bc9210b599fe648409db30ce2897f3723c5

Basically, this is used for setting up chat rooms, so all the chat-related tests are failing.

PR #1286, opened 2019-01-21
async create (connection middleware)

alimvedats:
~~Unlike action middleware, connection middleware is non-blocking and connection logic will continue as normal regardless of what you do in this type of middleware.~~

destroy method is async but why create method is sync?

### DIFF
--- a/classes/connection.js
+++ b/classes/connection.js
@@ -25,21 +25,66 @@ module.exports = class Connection {
    * @see ActionHero.Server
    *
    * @param  {Object} data The specifics of this connection
+   * @param  {Boolean} syncSetup The specifics of this connection initializing sync or async
    */
-  constructor (data) {
+  constructor (data, syncSetup = true) {
     // Only in files required by `index.js` do we need to delay the loading of the API object
     // This is due to cyclical require issues
     api = require('./../index.js').api
 
     this.setup(data)
 
-    api.connections.connections[this.id] = this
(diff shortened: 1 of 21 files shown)


alimvedats:
> Looks like a great start!
> 
> Here are some suggestions for next steps:
> 
> 1. ~~rebase from master so it would be possible to merge it in~~
> 2. ~~add a test showing off an async connection create.  Add an `await api.utils.sleep(1)` for a simple way to sleep.  This is probably the most important missing step~~
> 3. ~~update the docs to show that you need to call the new async method~~
> 4. ~~complete your TODOs for the docs~~
> 
> This is a breaking change!

👍done

classes/connection.js
```diff
@@ -25,21 +25,52 @@ module.exports = class Connection {
...
    */
-  constructor (data) {
+  constructor (data, syncSetup = true) {
     // Only in files required by `index.js` do we need to delay the loading of the API object
     // This is due to cyclical require issues
     api = require('./../index.js').api
 
     this.setup(data)
 
+    if (syncSetup) {
```
evantahler: What would be the case when you don't want syncSetup?

alimvedats: you right!
💡Also if I change the argument name to "callCreateMethods" I think this confusion will be disappear. 

```suggestion
    if (callCreateMethods) {
```

PR #1287, opened 2019-01-22
Adding strict transport security header as a default in web server config

grsmyth:
As requested in Slack, adding Strict Transport Security header in the default web http header section.

This means that when a request is made over https, it stores in your browser to always redirect you to the https site for the duration of the max-age even if you type in the http address.

### DIFF
--- a/config/servers/web.js
+++ b/config/servers/web.js
@@ -24,7 +24,8 @@ exports['default'] = {
           'X-Powered-By': api.config.general.serverName,
           'Access-Control-Allow-Origin': '*',
           'Access-Control-Allow-Methods': 'HEAD, GET, POST, PUT, PATCH, DELETE, OPTIONS, TRACE',
-          'Access-Control-Allow-Headers': 'Content-Type'
+          'Access-Control-Allow-Headers': 'Content-Type',
+          'Strict-Transport-Security': 'max-age=31536000; includeSubDomains'
         },
         // Route that actions will be served from; secondary route against this route will be treated as actions,
         //  IE: /api/?action=test == /api/test/


grsmyth:
So just some information on Strict Transport Security header.

When you visit a site that has this header on https. For example, 'https://www.google.com', this header tells your browser to lock this site to https. That means when you type 'google.com' in the browser, it will redirect you to 'https://www.google.com' for as long as the lifetime set in max-age (obviously reset when you revisit the site).

However this only kicks off when you visit the https site. If you visit http://www.google.com, nothing happens. It works like normal (well unless they build a redirect into the http site to make sure you always visit https but that's another system entirely...).

As such, there is no issue for anyone developing a non-secure site, it won't affect localhost development, and is only applicable to sites that have https capabilities. But really you shouldn't be developing non-https sites these days anyway. But that again is another topic!

crrobinson14:
I looked into this a little more. It seems this is correct, MDN has this to say:

> Note: The Strict-Transport-Security header is ignored by the browser when your site is accessed using HTTP; this is because an attacker may intercept HTTP connections and inject the header or remove it. When your site is accessed over HTTPS with no certificate errors, the browser knows your site is HTTPS capable and will honor the Strict-Transport-Security header.

It still makes me a little nervous, because the RFC itself doesn't appear to call this out. It seems like a behavior browser vendors could change in the future based on evolving security policies.

That said, I created a sample ActionHero project using the documented starter process at https://www.actionherojs.com/downloads/. I made this change manually to my local configuration then ran the service in both dev and production modes. I tested access to it via Safari, Firefox, and Chrome and encountered no issues. In all three browsers I inspected the network traffic to verify that the header was being sent, and follow up requests to verify that nothing broke. It looks good to me. I suppose if browser manufacturers do change their policies it's no different than any of the other past things they've done that developers have had to put up with - and not anything AH should be solely responsible for.

@grsmyth If this header is important enough to want to include by default it would be good to add a test case to verify that it's being sent. (There are code pathways where headers can be altered and we'd want to know nothing was nuking it.) If you have time please consider adding one for it. That said, I know from personal experience that adding barriers here stymies contributions, so hopefully Evan won't kill me but I'm going to merge this now. Thanks for your contribution!

PR #1288, opened 2019-01-24
Config merge empty objects fix, semantic undefined and null

shookie:
As discussed in #1283, the fixed merge behavior, where undefined values are ignored and null values trigger deletion of keys.

Also some more test cases.


evantahler:
Please update your code so it passes the linter -> https://circleci.com/gh/actionhero/actionhero/2385

Issue #1289, opened 2019-01-24
AH Task Tutorial Documentation


samking314:
# AH Task Tutorial Documentation
The AH Documentation for Tasks is not updated. Specifically, the page https://docs.actionherojs.com/tutorial-tasks.html
has an incorrect tutorial where it says that `initializers/node_schedule.js` should have in it
`initialize (api, next) {
    api.scheduledJobs = []
  }`
but this is wrong, because we're already requiring `api` from actionhero. Also, if you try running this code in your `initializers/node_schedule.js` you'll get an error saying `scheduleJobs` is undefined (because `api` is undefined because this is not the required `api`).

Instead, this documentation should say
`initialize () {
    api.scheduledJobs = []
  }`

- ActionHero Version: `19.0.4`
- Node.js Version: `10.14.1`
- Operating System: `(OSX)`

# Steps to reproduce your error
- Look at the tutorial under `initializers/node_schedule.js`

evantahler:
Thanks for pointing this out!   Are you able to update the documentation via a pull request?

samking314:
Sure can, Evan. Thanks for replying.

evantahler:
Solved by https://github.com/actionhero/actionhero/pull/1290

PR #1290, opened 2019-01-25
fixed initializers/node_schedule.js example

samking314:
this fixes the Github issue I raised in #1289


samking314:
This is curious. I am not sure why it's not passing. Either way, if you're using AH v19 and node v10 then this fixes the tutorial. Although, there could be something going on behind the scenes of which I am unaware.

evantahler:
Thanks for the update!  I'll check on CircleCI.  Sometimes the test times out, or takes a bit too long to start.

PR #1305, opened 2019-05-28
Support for custom file globs when loading tasks, servers, and actions.

jamespedid:
- Adds support for a custom file glob to load action hero actions, tasks, or servers.


config/api.js
```diff
@@ -57,6 +57,9 @@ exports['default'] = {
         'plugin': [path.join(__dirname, '/../node_modules')],
         'locale': [path.join(__dirname, '/../locales')]
       },
+      // file glob that will be considered when looking at the paths for actionhero resources
+      // relative to paths defined above
+      fileGlob: '**/*.js',
```
jamespedid: For a first pass at this, I have added a single string property for fileglobs that defaults to **/*.js, and this applies for each actions, tasks, and servers. Currently no logic exists that allows for a separate glob for each to exist. (Perhaps overload this with ability to support an object with action, task, and server keys to have their own globs?)

initializers/tasks.js
```diff
@@ -547,15 +547,15 @@ module.exports = class Tasks extends ActionHero.Initializer {
 
     api.tasks.loadTasks = (reload) => {
       api.config.general.paths.task.forEach((p) => {
-        glob.sync(path.join(p, '**', '*.js')).forEach((f) => {
+        glob.sync(path.join(p, api.config.fileGlob)).forEach((f) => {
```
jamespedid: Right now we assume that because it exists in the config file provided by actionhero that it is exists. Do we want to have a little bit safer logic here?

initializers/actions.js
```diff
@@ -134,14 +134,14 @@ module.exports = class Actions extends ActionHero.Initializer {
...
-      let files = glob.sync(path.join(p, '**', '*.js'))
+      let files = glob.sync(path.join(p, api.config.fileGlob))
       for (let j in files) { await api.actions.loadFile(files[j]) }
     }
 
     for (let pluginName in api.config.plugins) {
       if (api.config.plugins[pluginName].actions !== false) {
         let pluginPath = api.config.plugins[pluginName].path
-        let files = glob.sync(path.join(pluginPath, 'actions', '**', '*.js'))
+        let files = glob.sync(path.join(pluginPath, 'actions', api.config.fileGlob))
```
jamespedid: unsure if this should also apply to plugins. Should plugins have their own settings for this kind of thing?

evantahler: This is going to get really tricky.  Right now, we can assume that all plugins publish *js files (even if they are compiled from something else).  If you were to change this, you open up the possibility that some plugins pulish *js files and others don't.  This seems to imply that each plugin needs to then define the types of files it produces... perhaps in package.json?

evantahler: I think you need to assume that plugins still only publish *.js files.  We can't assume that all actionhero projects would have the proper engine/transpiler to use typescript files, for example

config/api.js
evantahler: Perhaps the thing to do might be to modify the `paths` rather than add a new key.  You would then be able to define some typescript paths, and some js paths:

```js
api.config.general.paths.actions = [
  path.join(__dirname, '/../actions/**/*.js'),
  path.join(__dirname, '/../actions/**/*.ts')
]
```

The default would be to use *.js files.
This would be a breaking change and trigger the release of actionhero v20

jamespedid: I like the idea for this change. It fits into the existing config already.

evantahler:
I think this will be solved by https://github.com/actionhero/actionhero/pull/1356

Issue #1308, opened 2019-06-07
Exception in periodic task stops this task queue


dasagan:
# Name of Issue
Exception in periodic task stops this task queue 

- ActionHero Version: `19.1.4`
- Node.js Version: `8.11.3`
- Operating System: `(Windows 10)`

# Steps to reproduce your error
To create test-task.js with content:
```javascript
const {api, Task} = require('actionhero')

exports.SayHello = class SayHello extends Task {
  constructor () {
    super()
    this.name = 'sayHello'
    this.description = 'I say hello'
    this.frequency = 1000
    this.queue = 'low'
    this.middleware = []
  }

  async run () {
    api.log("hello",'info');
    throw 'exception from SayHello periodic task';
  }
}
```

# Steps to fix error
Old snippet in actionhero/initializers/tasks.js:
```javascript
      return {
        plugins: plugins,
        pluginOptions: pluginOptions,
        perform: async function () {
          let combinedArgs = [].concat(Array.prototype.slice.call(arguments))
          combinedArgs.push(this)
          let response = await task.run.apply(task, combinedArgs)
          await api.tasks.enqueueRecurrentTask(taskName)
          return response
        }
      }
    }
```
to replace by new:
```javascript
      return {
        plugins: plugins,
        pluginOptions: pluginOptions,
        perform: async function () {
          let combinedArgs = [].concat(Array.prototype.slice.call(arguments))
          combinedArgs.push(this)
          let response = null;
          try{
          	response = await task.run.apply(task, combinedArgs)
          } catch (error) {
          	throw error
          }
          finally {
          	await api.tasks.enqueueRecurrentTask(taskName)
          }
          return response
        }
      }
    }
```

evantahler:
Thanks!

I don't know if this is a bug per-se, but was a design choice to not keep re-running a crashing job.  That said, perhaps this belongs as an option to re-run crashing jobs or not? Are you able to make a Pull request with your changes we can discuss?

evantahler:
Amending my note above, it might be cool to have the new `api.config.tasks.reEnqueuePeriodicTasksIfException` (or whatever you call the option) as default enabled for the `development` environment, and then off for all the others?

samking314:
I feel like this is more useful for very specific jobs. For instance, if I have a job that I know might throw an error initially but will return successfully at some point over a bounded period of time then it might be useful for that job to keep crashing until it returns successfully(although now that I write that down it seems like a very unlikely scenario). 
[...]

PR #1309, opened 2019-06-10
reEnqueuePeriodicTaskIfException

dasagan:
Fix: Exception in periodic task stops this task queue
New task option: `reEnqueuePeriodicTaskIfException`

Solves https://github.com/actionhero/actionhero/issues/1308


initializers/tasks.js
```diff
@@ -147,8 +147,16 @@ module.exports = class Tasks extends ActionHero.Initializer {
         perform: async function () {
           let combinedArgs = [].concat(Array.prototype.slice.call(arguments))
           combinedArgs.push(this)
-          let response = await task.run.apply(task, combinedArgs)
-          await api.tasks.enqueueRecurrentTask(taskName)
+          let response = null;
```
evantahler: This fails the linter (standard.js).  You can see why here https://circleci.com/gh/actionhero/actionhero/2949?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

You can fix this automatically with `standard --fix`

evantahler:
Thank you for your work @dasagan!  I've been asking a lot of folks for their opinion on this, and it turns out that we should make the default behavior to re-enqueue if there is a failure.  It's more surprising that recurring tasks... aren't recurring to folks.  

I'm going to merge your work in, but change the option around so the added option is actually to *not* to re-enqueue failing tasks... and by default they will be.

PR #1311, opened 2019-06-12
Boot loader for actionhero

chimmelb:
Top level (optional) file for actionhero CLI that loads before everything else. The "newrelic" or "dotenv" use-case. Added tests too.

See bin/templates/boot.js for the example/empty version. 

Where should this be mentioned in the docs? I will update there too.


bin/actionhero
```diff
@@ -18,6 +12,19 @@ if (process.env.projectRoot) {
 } else {
   projectRoot = path.normalize(process.cwd())
 }
+try {
+  const boot = require(projectRoot + path.sep + 'boot.js')
+  boot()
```
evantahler: ... can this be async?

bin/templates/boot.js
```diff
@@ -0,0 +1,11 @@
+'use strict'
+/* This file will load, be `require()`d, first in the actionhero binary.
+   Anything that needed loaded or run BEFORE actionhero starts can be done here.
+
+   This file returns a single method that is run immediately after loading.
+*/
+
+// include high-level requires here.
+
+module.exports = function () {
```
evantahler: ... `async function`

chimmelb:
Loading environments from remote or I suppose system-level checks before starting AH sounds cool.

I made the change that wraps all of `bin/actionhero` in something like this (thus the larger code difference)
```
(async () => {
 // entire rest of the file
})()
```
Updated the CLI test to include an await as well, before testing the custom "BOOTING" message.  Hopefully that doesn't change too much, can add more tests or change back to sync loading.

evantahler:
Thanks @chimmelb.  It looks like you've got some merge conflicts to clean up before this can be merged in

evantahler:
I'll just make a merge commit!

chimmelb:
Thank you handling that merge and accepting the PR. I'll watch for the next actionhero release!

PR #1313, opened 2019-06-28
Implemented config option to force "full" reload on action/task change

### LINKED ISSUE #1275: Change module-reload to full server reload
# Change module-reload to full server reload
This is an RFP for reworking the reload on file changes to perform a full server reload rather than reloading modules and hot-swapping.

It would also be nice to expand the "watch and reload" capability to plugins to avoid duplicating functionality in each plugin (ie: https://github.com/actionhero/ah-sequelize-plugin/blob/d0491b7f8e9e14289de5141dabd9264308f602ad/classes/sequelize.js#L63-L69)

aurasalexander:
Resolves #1275

Opted to make this configurable with the default being the old behavior.
I tested the new code paths, but didn't create new test cases (primarily due to my doubts that this can be automatically tested at all).

### DIFF
--- a/config/api.js
+++ b/config/api.js
@@ -20,8 +20,13 @@ exports['default'] = {
       lockPrefix: 'actionhero:lock:',
       // how long will a lock last before it exipres (ms)?
       lockDuration: 1000 * 10, // 10 seconds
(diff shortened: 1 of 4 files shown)


evantahler:
Great!

PR #1314, opened 2019-06-29
Refactor api.log loglevel severity

aurasalexander:
Preface:
ActionHero uses the winston syslog loglevels.
The syslog protocol (RFC 5424) knows 8 severities levels:

Code | Level | Description
-- | -- | --
0 | Emergency | system is unusable
1 | Alert | action must be taken immediately
2 | Critical | critical conditions
3 | Error | error conditions
4 | Warning | warning conditions
5 | Notice | normal but significant condition
6 | Informational | informational messages
7 | Debug | debug-level messages

Issue:
I noticed that ActionHero logs some messages with severities worthy of discussion. For example, when I define an error logger in winston, the error log will grow with `*** ActionHero Started ***` and `*** ActionHero Stopped ***` log entries because they are logged with severity `alert`. I think I don't have to take immediate action just because ActionHero started and I feel like it shouldn't be logged in an error log. I took the oportunity and went through all api.log calls to check on their severity.

I would purpose the following changes, each of them entirely open for discussion:

File | Log cause | Previous level | Proposed level | Reason
-- | -- | -- | -- | --
`classes/process.js` | Initializer exception | warning | emerg | ActionHero will fail to start
^ | ActionHero starting | notice | info | Only indicating begin of start process
^ | ActionHero start/restart/stopped | alert | notice | Normal condition with significance
`config/redis.js` | Redis connection first retry | error | alert | If redis connection retries continue to fail (if no immediate action is taken), ActionHero will fail to start 
`initializers/actions.js` | Action reloaded | debug | info | Should get informed about reloads (not initial loads) to make obvious which code version is currently loaded in developmentMode
`initializers/redis.js` | Redis connection further retries | error | alert | If redis connection retries continue to fail (if no immediate action is taken), ActionHero will fail to start 
`initializers/resque.js` | Node elected as new resque master | info | notice | Normal but significant condition
`initializers/specHelper.js` | Loading testServer | warning | info | Nothing to warn about, informational character
`initializers/tasks.js` | Tasks reloaded | debug | info | Should get informed about reloads (not initial loads) to make obvious which code version is currently loaded in developmentMode

### DIFF
--- a/classes/process.js
+++ b/classes/process.js
@@ -140,7 +140,7 @@ module.exports = class Process {
           } catch (error) {
             let message = `Exception occured in initializer \`${initializer.name}\` during load`
             try {
-              api.log(message, 'warning', error.toString())
+              api.log(message, 'emerg', error.toString())
             } catch (_error) {
               console.error(message)
             }
@@ -156,7 +156,7 @@ module.exports = class Process {
             await initializer.start()
             api.log(`Started initializer: ${initializer.name}`, 'debug', file)
           } catch (error) {
-            api.log(`Exception occured in initializer: ${initializer.name} during start`, 'warning', error.toString())
+            api.log(`Exception occured in initializer \`${initializer.name}\` during start`, 'emerg', error.toString())
             throw error
           }
         }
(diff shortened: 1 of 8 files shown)


evantahler:
Awesome work, thank you!

PR #1321, opened 2019-07-15
Changes for Standard v13

evantahler:


evantahler:
The red tests are testing an old version of the code from the CLI with new standard, resulting in the linter failures.  This will be resolved on the next release

Issue #1005, opened 2017-02-27
Roadmap: Better spec helpers


evantahler:
- ActionHero Version: all
- Node.js Version: all
- Operating System: all

When you generate a task or action, a test for that new file should also be generated, and you should be able to run that test with no additional setup

---

- [x] Generating an action should generate a related test
- [x] we can test that a task was enqueued (normal queue)
- [x] we can test that a task was enqueued (future queue)

evantahler:
This will require us to make some choices about test runners (mocha) and assertion tools (chai), and also how to book the server in test mode.  This "booting" process should be part of core.

evantahler:
Task enqueuing test helpers added via https://github.com/actionhero/actionhero/pull/1331

gcoonrod:
Anything else @evantahler?

evantahler:
I think that's it!

PR #1332, opened 2019-09-03
create tests for actions and tasks when generating via CLI

evantahler:
Completes https://github.com/actionhero/actionhero/issues/1005

When generating a new `action` or `task` via the CLI, a template test will also now be creates in `__tests__`.  This will help encourage better testing of your actionhero projects.  Of course, you can always delete the test files if you don't want it.

Adds `api.config.general.paths.test` as setting in `./config/api.js` which defaults to `[path.join(__dirname, '/../__tests__')]`


evantahler:
This is a breaking change due to the new, required `api.config.general.paths.test` config option

PR #1334, opened 2019-09-05
feature: add custom http status code with error object

piyushagwl007:
Currently, we have to add the custom error response code like this:
`data.connection.rawConnection.responseHttpCode = 404`     
We can use the error.code that already exists in the `error object`

Now in action, you can do something like this :
```js
const raiseNotFound = <your-condition>
if(raiseNotFound) {
  const notFoundError = new Error(<msg>)
  notFoundError.code = 404
   throw notFoundError
}
```

### DIFF
--- a/__tests__/servers/web/web.js
+++ b/__tests__/servers/web/web.js
@@ -677,25 +677,6 @@ describe('Server: Web', () => {
   describe('http returnErrorCodes true', () => {
     let orig
     beforeAll(() => {
-      function SuspiciousActivity (message, extraFields) {
-        Error.call(this)
-
-        this.name = this.constructor.name
-        this.message = message || '402: Suspicious activity detected'
-        this.responseHttpCode = 402
-        Object.assign(this, extraFields || {})
-      }
-
-      class NotFoundError extends Error {
-        constructor (message, extraFields) {
-          super()
-          this.name = this.constructor.name
-          this.message = message || '404: Not FOund'
-          this.responseHttpCode = 404
-          Object.assign(this, extraFields || {})
-        }
-      }
-
       orig = api.config.servers.web.returnErrorCodes
       api.config.servers.web.returnErrorCodes = true
 
@@ -719,7 +700,9 @@ describe('Server: Web', () => {
               const validQueryFilters = ['test', 'search']
               const validQueryParam = validQueryFilters.indexOf(data.params.query) > -1
               if (!validQueryParam) {
-                throw new NotFoundError(`404: Filter '${data.params.query}' not found `)
+                const notFoundError = new Error(`404: Filter '${data.params.query}' not found `)
+                notFoundError.code = 404
+                throw notFoundError
               }
             }
             const hasRandomKey = !!data.params.randomKey
@@ -727,7 +710,9 @@ describe('Server: Web', () => {
               const validRandomKeys = ['key1', 'key2', 'key3']
               const validRandomKey = validRandomKeys.indexOf(data.params.randomKey) > -1
               if (!validRandomKey) {
-                throw new SuspiciousActivity(`402: Suspicious Activity detected with key ${data.params.randomKey}`)
+                const suspiciousError = new Error(`402: Suspicious Activity detected with key ${data.params.randomKey}`)
+                suspiciousError.code = 402
+                throw suspiciousError
               }
             }
             data.response.good = true
--- a/__tests__/servers/web/web.js
+++ b/__tests__/servers/web/web.js
@@ -677,6 +677,25 @@ describe('Server: Web', () => {
   describe('http returnErrorCodes true', () => {
     let orig
     beforeAll(() => {
+      function SuspiciousActivity (message, extraFields) {
+        Error.call(this)
+
+        this.name = this.constructor.name
+        this.message = message || '402: Suspicious activity detected'
+        this.responseHttpCode = 402
+        Object.assign(this, extraFields || {})
+      }
+
+      class NotFoundError extends Error {
+        constructor (message, extraFields) {
+          super()
+          this.name = this.constructor.name
+          this.message = message || '404: Not FOund'
+          this.responseHttpCode = 404
+          Object.assign(this, extraFields || {})
+        }
+      }
+
       orig = api.config.servers.web.returnErrorCodes
       api.config.servers.web.returnErrorCodes = true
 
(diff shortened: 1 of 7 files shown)


classes/actionProcessor.js
```diff
@@ -98,6 +98,11 @@ module.exports = class ActionProcessor {
     }
 
     if (error) {
+      const hasCustomResponseCode = error.responseHttpCode && this.connection.rawConnection
```
evantahler: I think the `actionProcessor`, is the wrong place for this logic.  Every type of connection (websocket, socket, and custom ones) will have a `connection.rawConnection`).  This might belong with the other response code logic for web server @ https://github.com/actionhero/actionhero/blob/master/servers/web.js#L349-L361

config/servers/web.js
```diff
@@ -80,6 +80,44 @@ exports.default = {
...
+         *        super()
+         *        this.name = this.constructor.name
+         *        this.message = message || '404: Not FOund'
+         *        this.responseHttpCode = 404
+         *        Object.assign(this, extraFields || {})
+         *         }
+         *       }
+         *
+         *       function SuspiciousActivity (message, extraFields) {
+         *          Error.call(this)
```
evantahler: What's `this` in this function?

It might be clearer to:
```js
const error = new Error('402: Suspicious activity detected')
error.code = 402
return error
```

evantahler: Node internally uses `error.code` which means that it's a first-class property we can start relying on https://nodejs.org/api/errors.html#errors_error_code

evantahler: This note probably doesn't belong in the config file for the web server, but perhaps in the Tutorial for either Actions or the Web Server

piyushagwl007: > It might be clearer to create the error directly in the `run` method of the action rather than using a generator/factory:
Ohh!! Makes sense, will do that


> This note probably doesn't belong in the config file for the webserver, but perhaps in the Tutorial for either Actions or the Web Server

Can you guide me on where to add this so that this will come inside the Tutorial for Actions or Web Server?

classes/actionProcessor.js
piyushagwl007: I see your point, yes moving it ro the webserver logic makes more sense.
 will move this logic to the required file. :smile:

config/servers/web.js
piyushagwl007: Should I add them in the [action tutorial](https://github.com/actionhero/actionhero/blob/master/tutorials/actions.md) OR [web-server tutorial](https://github.com/actionhero/actionhero/blob/master/tutorials/web-server.md)? 

Adding them to the webserver tutorial makes more sense to me, thoughts?

__tests__/servers/web/web.js
```diff
@@ -686,15 +705,32 @@ describe('Server: Web', () => {
...
             }
+            const hasQueryParam = !!data.params.query
+            if (hasQueryParam) {
+              const validQueryFilters = ['test', 'search']
+              const validQueryParam = validQueryFilters.indexOf(data.params.query) > -1
+              if (!validQueryParam) {
+                throw new NotFoundError(`404: Filter '${data.params.query}' not found `)
+              }
+            }
+            const hasRandomKey = !!data.params.randomKey
```
gcoonrod: I've been out of touch with JS for a bit. What does `!!data.params.randomkey` do?

piyushagwl007: If `randomKey` is not passed in the request, so it will be `undefined`, so `!undefined` will be true and doing `!!undefined` will become false which means that the variable is not sent.    

If `randomKey` is set it will return true for `!randomKey` and doing `!!randomKey` will return true.
    
I wanted to not do this kind of thing as I know it seems confusing, but I was unable to find any kind of `isEmpty` sort-of Util but was unable to find, so used this type of thing.

Should I remove this and create an `isEmpty` sort-of Util and go ahead with that?

gcoonrod: I believe the typical convention is to use `const hasRandomKey = data.params && data.params.randomKey`. Has that changed recently?

piyushagwl007: You are absolutely right, doing `const hasRandomKey = data.params && data.params.randomKey`  would produce the same result, but then `hasRandomKey` will contain the value of `data.params.randomKey` not a `boolean`, I thought that since the variable name is `hasRandomKey` it represents a `boolean`, plus that makes easy to read the if condition rather than predicting what it will do. Is this a good way to go?

servers/web.js
```diff
@@ -348,7 +348,11 @@ module.exports = class WebServer extends ActionHero.Server {
 
     if (data.response.error) {
       if (this.config.returnErrorCodes === true && data.connection.rawConnection.responseHttpCode === 200) {
-        if (data.actionStatus === 'unknown_action') {
+        const customErrorCode = parseInt(data.response.error.code, 10)
+        const isValidCustomResponseCode = customErrorCode >= 100 && customErrorCode < 600
```
evantahler: I love this check! 
Please include a test for this, showing that an error code of `999` won't make it to client.

tutorials/web-server.md
```diff
@@ -151,6 +151,29 @@ exports['default'] = {
         },
         // When true, returnErrorCodes will modify the response header for http(s) clients if connection.error is not null.
         // You can also set connection.rawConnection.responseHttpCode to specify a code per request.
+                /**
+         *  To create custom Error objects with custom response code you need to add the your response status code in the error object eg:
```
evantahler: I think a little more needs to be said here about how "... using `error.code` will overwrite the default behavior described above."

piyushagwl007: Sure

__tests__/servers/web/web.js
```diff
@@ -743,6 +769,58 @@ describe('Server: Web', () => {
...
+        expect(receivedBody.good).toEqual(true)
+      })
+
+      test('should not work for 999 status code set using custom error and default error code, 400 is thrown', async () => {
+        try {
+          await request.post(url + '/api/statusTestAction', { form: { key: 'value', randomKey: 'expired-key' } })
+          throw new Error('should not get here')
+        } catch (error) {
+          expect(error.statusCode).not.toEqual(999)
+          expect(error.statusCode).toEqual(400)
```
piyushagwl007: @evantahler Added the required test :smile:

PR #1338, opened 2019-09-25
Always prefer actionhero in node_modules

evantahler:
Related to https://github.com/actionhero/ah-swagger-plugin/pull/6

If we change how Actionhero loads itself to *always* prefer components in `node_modules` relative to the `cwd` (current working directory):

**Pros**
* There is no change to the behavior or a normal actionhero project
* Globally installed Actionhero (`npm install -g actionhero`) is now resistant to minor version changes between projects
* Allows plugins to install a version of actionhero as a `devDependancy`, and for us to boot the project and not have conflicting paths to source actionhero from.  You can now just `npm start` (`./node_modules/.bin/actionhero`) in your plugins to start a server to run your plugins

**Cons**
* If you start actoinhero from a dameon (like `forever`) without first `cd`-ing to the proper directory first things might not work? 
* Slightly slower boot time (as we are checking if files exist before trying to load them)
* Gets weird with yarn/lerna workspaces
* ? -> Add some in the comments!

### DIFF
--- a/config/api.js
+++ b/config/api.js
@@ -51,17 +51,17 @@ exports.default = {
       cliIncludeInternal: true,
       // configuration for your actionhero project structure
       paths: {
-        action: [path.join(__dirname, '/../actions')],
-        task: [path.join(__dirname, '/../tasks')],
-        public: [path.join(__dirname, '/../public')],
-        pid: [path.join(__dirname, '/../pids')],
-        log: [path.join(__dirname, '/../log')],
-        server: [path.join(__dirname, '/../servers')],
-        cli: [path.join(__dirname, '/../bin')],
-        initializer: [path.join(__dirname, '/../initializers')],
-        plugin: [path.join(__dirname, '/../node_modules')],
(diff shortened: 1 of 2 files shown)


evantahler:
I think we should mark this a breaking change. The boot behavior of tools like forever might not work if you don’t CD into the directory now

piyushagwl007:
@evantahler So by introducing this we won't have to do `require.resolve` in plugins? As I did [here](https://github.com/actionhero/ah-swagger-plugin/blob/bb72edf171f1ef12ba20a1982292eeabade40cfc/initializers/ah-swagger-plugin.js#L2-L6)
``` js
const directoriesToSearch = [process.cwd()]
const actionHeroInstanceIndexPath = require.resolve('actionhero', {
  paths: directoriesToSearch
})
const { Initializer, api } = require(actionHeroInstanceIndexPath)

```

evantahler:
@piyushagwl007 I *think* that's correct.  I'll merge this in, and you can try it out!

Issue #1343, opened 2019-10-17
It should be more clear that config.servers.web.secure is about SSL termination and not redirection


evantahler:
from https://actionherojs.slack.com/archives/C04EVSUSD/p1571324381069900

1. Clarify that https://github.com/actionhero/actionhero/blob/master/config/servers/web.js#L11 is talking about SSL termination locally in the app, and not if the host headers require `https`.  This should be done in the config file comments and in https://github.com/actionhero/actionhero/blob/master/tutorials/web-server.md
2. Clarify that ALLOWED_HOSTS is meant to be used fro redirects, including switching from http to https (https://github.com/actionhero/actionhero/blob/master/config/servers/web.js#L16)  This should be done in the config file comments and in https://github.com/actionhero/actionhero/blob/master/tutorials/web-server.md and also https://github.com/actionhero/actionhero/blob/master/tutorials/production-notes.md

evantahler:
solved by https://github.com/actionhero/actionhero/pull/1345

PR #1345, opened 2019-10-21
add note about what api.config.servers.web.secure does

evantahler:


evantahler:
solves https://github.com/actionhero/actionhero/issues/1343

PR #1365, opened 2019-11-21
Combined utils

evantahler:


evantahler:
@gcoonrod Here's a good example of why the utils should be initially sourced in separate files: https://github.com/actionhero/actionhero/pull/1365/commits/be72b6fa781fc41d08a079a13f52d97358ad7ef4

We want to use the utils when the project doesn't exist yet.  Some of the utils rely on `config`, and if we were to load them all, it would crash.  This lets us only load the ones we need.

evantahler:
Thanks @gcoonrod!

PR #1366, opened 2019-11-23
load files from plugins in `SRC` or `DIST`

evantahler:


evantahler:
I imagine in these paths will need to be variable in the future...

PR #1367, opened 2019-11-23
Adding polyfill for legacy api features.

gcoonrod:
Poly fill the new `api` object with all of the old functions.


__tests__/core/polyfill.ts
```diff
@@ -0,0 +1,28 @@
...
+            api.utils,
+            api.cache,
+            api.tasks,
+            api.actions,
+            api.resque
+        ].forEach(item => {
+            expect(item).toBeInstanceOf(Object);
+        });
+
+        expect(api.log).toBeInstanceOf(Function);
```
evantahler: Perhaps it will be useful to `call` one of the methods (like `api.cache.load` and `api.cache.save`) to make sure they do something?

src/initializers/polyfill.ts
```diff
@@ -0,0 +1,38 @@
...
+    this.loadPriority = 999;
+  }
+
+  async initialize(config) {
+    const development =
+      !config.process.env || config.process.env === "development";
+    const api = apiExports.api;
+
+    if (development) {
+      apiExports.log("Polyfilling [api] object", "info");
```
evantahler: Awesome! A message only in development.  I would add more to it like... "...using `api.cache`, etc will soon be depreciated in favor of importing those namespaces from Actionhero directly"

src/initializers/polyfill.ts
```diff
@@ -0,0 +1,38 @@
...
+
+      if (key === "api" || api[key]) {
+        continue;
+      }
+
+      api[key] = apiExports[key];
+    }
+
+    if (development) {
+      apiExports.api.log("Polyfilled [api] object", "info");
```
evantahler: I think you are logging a warning twice? Once is probably sufficient...   Or, this message could be moved to `debug`

evantahler:
Oh! What about `api.config`?

__tests__/core/polyfill.ts
gcoonrod: done

src/initializers/polyfill.ts
gcoonrod: done

src/initializers/polyfill.ts
gcoonrod: done

evantahler:
Great Work!

Issue #1375, opened 2019-12-02
Breaking change in API versioning in v21


inian:
- ActionHero Version: `21.x.x`

I see that API versioning has been changed from /api/v1/showDocumentation to /api/showDocumentation?apiVersion=1

This should be documented with the list of other breaking changes [here](https://www.actionherojs.com/tutorials/typescript)

Separately, I liked the api version being in the path rather than in the query string. Was there a specific reason why the version was moved to the query string? 

Thanks in advance..

evantahler:
That's certainly not an intentional change! 

It looks like a routes file with 
```ts
export const DEFAULT = {
  routes: config => {
    return {
      get: [{ path: "/:apiVersion/randomNumber", action: "randomNumber" }]
    };
  }
};
```
works for `/api/1/randomNumber`

but with a prefix

```ts
export const DEFAULT = {
  routes: config => {
    return {
      get: [{ path: "/v:apiVersion/randomNumber", action: "randomNumber" }]
    };
  }
};

```
it doesn't work, ie: `/api/v1/randomNumber`.  Are you able to work on this bug?

inian:
Previously, I used to set `this.action = "v1"` directly in the constructor of the action. Now looks like I can't set it to v1 directly since `this.version` is a number. 

So, is the expected behaviour to add a version number in `this.action` and a prefix (like "v") in the routes.ts file? This prefix mechanism wasn't supported in v20. 

We can either support adding a prefix for apiVersion or make apiVersion a string.

evantahler:
Ah! That makes much more sense!

We never had a test case ensuring explicitly that action versions could be strings... we do now via https://github.com/actionhero/actionhero/pull/1376.

Thank you for testing this new version!

PR #1376, opened 2019-12-02
Action Versions can be numbers or strings

evantahler:


evantahler:
closes https://github.com/actionhero/actionhero/issues/1375

PR #1381, opened 2019-12-12
Use --no-deps in ts-node-dev

inian:
Right now ts-node-dev watches all direct dependencies in node_modules which is unnecessary. [Adding --no-deps](https://github.com/whitecolor/ts-node-dev/issues/51#issuecomment-512096067) makes server restarts *significantly faster*.  People shouldn't be modifying node_modules directly anyway and server restarts anyway when package.json changes, so I think this is a good thing to have by default.

### DIFF
--- a/templates/package.json.template
+++ b/templates/package.json.template
@@ -24,7 +24,7 @@
   },
   "scripts": {
     "postinstall": "npm run build",
-    "dev": "ts-node-dev --transpile-only ./node_modules/.bin/actionhero start",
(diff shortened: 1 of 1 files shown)


evantahler:
That makes a ~4 second difference for me! That's wonderful!

evantahler:
I added that to the "project" `package.json` as well via cad0adb95afcad818b27617c85b1bc9d4cfce243

PR #1395, opened 2020-01-27
Adding 2020 Developer survey link

gcoonrod:


gcoonrod:
Hmm, all those commits seem odd...

evantahler:
Yeah... but I'll squash them so they will go away.  Looks like we've had some merge commits since you last pulled.

jamespedid:
Not sure where else to put this, but the survey forced me to pick that I was using websockets or tcp sockets, even though I do not use either one.

evantahler:
Oops... Thanks!
Screen Shot 2020-02-18 at 1 56 12 PM

PR #1404, opened 2020-02-16
server.ts (remove `start` & `start cluster` binaries, along with boot.ts)

evantahler:
This PR removes the built-in `actionhero start` and `actionhero start cluster` commands, and adds a single `server.ts` entrypoint for your applications.  We are also removing support for `boot.ts|js`.

For years, Actionhero has shipped with a robust runtime solution, handling graceful restarts, clustering, and more... however. as the node.js ecosystem has matured, the community has:
1. Created excellent process management tools, like [PM2](https://pm2.keymetrics.io/) which handle running and monitoring your processes, including integration with OS process managers
2. Moved to serverless/PASS/Docker deployments which prefer to mange the running of the application directly.

To better integrate with the above, Actionhero will now run from a new `server.ts` file, which will be required in your project @ `/src/server.ts`:

```ts
#!/usr/bin/env node
import { Process } from "actionhero";

// load any custom code, configure the env, as needed

async function main() {
  // create a new actionhero process
  const app = new Process();

  // handle unix signals and uncaught exceptions & rejections
  app.registerProcessSignals();

  // start the app!
  // you can pass custom configuration to the process as needed
  await app.start();
}

main();
```

When developing, there is no change to your workflow: 
*  `npm run dev` will continue to hot-compile your typescript code and run it.  
* `npm run build` will compile your TS into JS
* `npm start` will now run your `./dist/server.js` instead of running `./node_modules/.bin/actoinhero`

This change should also make it easier to distribute your actionhero projects as you can modify a base config collection from this file, and you can use tools like [pkg](https://github.com/zeit/pkg) to compile your projects.

Now that we have this single entrypoint for your applications, we no longer need `boot.js|ts`. You can now directly run any setup code you need, modify the environment, etc directly in `server.ts`.


evantahler:
And if you are curious what `registerProcessSignals` looks like:

```ts
/**
   * Register listeners for process signals and uncaught exceptions & rejections.
   * Try to gracefully shut down when signaled to do so
   */
  registerProcessSignals() {
    function awaitHardStop() {
      const timeout = process.env.ACTIONHERO_SHUTDOWN_TIMEOUT
        ? parseInt(process.env.ACTIONHERO_SHUTDOWN_TIMEOUT)
        : 1000 * 30;
      return setTimeout(() => {
        console.error(
          `Process did not terminate within ${timeout}ms. Stopping now!`
        );
        process.nextTick(process.exit(1));
      }, timeout);
    }

    // handle errors & rejections
    process.on("uncaughtException", (error: Error) => {
      log(error.stack, "fatal");
      process.nextTick(process.exit(1));
    });

    process.on("unhandledRejection", (rejection: Error) => {
      log(rejection.stack, "fatal");
      process.nextTick(process.exit(1));
    });

    // handle signals
    process.on("SIGINT", async () => {
      log(`[ SIGNAL ] - SIGINT`, "notice");
      let timer = awaitHardStop();
      await this.stop();
      clearTimeout(timer);
    });

    process.on("SIGTERM", async () => {
      log(`[ SIGNAL ] - SIGTERM`, "notice");
      let timer = awaitHardStop();
      await this.stop();
      clearTimeout(timer);
    });

    process.on("SIGUSR2", async () => {
      log(`[ SIGNAL ] - SIGUSR2`, "notice");
      let timer = awaitHardStop();
      await this.restart();
      clearTimeout(timer);
    });
  }
```

jamespedid:
This hopefully should make using actionhero with container technologies simpler.

PR #1411, opened 2020-02-20
Tasks have inputs which are validated when enqueued

evantahler:
Like Actions, Tasks can now define the inputs they expect.   This is optional.

Inputs can be:
* required
* default
* validated

Unlike actions, we don’t need a formatter, as the inputs should already be of the proper type, coming from the server.  We can check the inputs at ‘enqueue’ rather than at runtime. This will ensure that no task without the required inputs is enqueued.

For example, with this task: 
```ts
import { Task } from "actionhero";

class SendWelcomeEmail extends Task {
  constructor() {
    super();
    this.name = "sendWelcomeEmail";
    this.description = "send a new user a welcome email";
    this.queue = "email";
    this.frequency = 0;
    this.inputs = {
      email: { required: true },
      template: { required: true, default: "welcome-email-en" }
    };
  }

  async run(params) {
    // send the email
  }
}
```

* `await task.enqueue('sendWelcomeEmail')` would throw an error, as "email" is a required input
* `await task.enqueue('sendWelcomeEmail', {email: 'evan@actionherojs.com'})` would be ok, and in the task `params.template` would be set to `welcome-email-en`, per the defaults.

There are also validators you can use, and like actions, you can throw a custom error or return false to prevent the task from being enqueued, ie:

```ts
import { Task } from "actionhero";

function emailValidator(p) {
  if (p.indexOf("@") < 0) {
    throw new Error("that is not an email address");
  }
}

class SendWelcomeEmail extends Task {
  constructor() {
    super();
    this.name = "sendWelcomeEmail";
    this.description = "send a new user a welcome email";
    this.queue = "email";
    this.frequency = 0;
    this.inputs = {
      email: { required: true, validator: emailValidator },
      template: { required: true, default: "welcome-email-en" }
    };
  }

  async run(params) {
    // send the email
  }
}
```
* `await task.enqueue('sendWelcomeEmail', {email: 'someone'})` would throw an error, as the email is missing the `"@"`
* `await task.enqueue('sendWelcomeEmail', {email: 'evan@actionherojs.com'})` would be ok

### DIFF
--- a/__tests__/core/tasks/tasks.ts
+++ b/__tests__/core/tasks/tasks.ts
@@ -67,22 +67,66 @@ describe("Core: Tasks", () => {
       }
     }
 
+    class TaskWithInputs extends Task {
+      constructor() {
+        super();
+        this.name = "taskWithInputs";
+        this.description = "task: taskWithInputs";
+        this.queue = queue;
+        this.frequency = 0;
+        this.inputs = {
+          a: { required: true, default: 1 },
+          b: {
+            required: true,
+            default: () => {
+              return 2;
+            }
+          },
+          c: {
+            required: true,
+            validator: p => {
+              if (p !== 3) {
+                throw new Error("nope");
+              }
+            }
+          },
+          d: {
+            required: true,
+            validator: p => {
+              if (p !== 4) {
+                return false;
+              }
+            }
+          }
+        };
+      }
+
+      async run(params) {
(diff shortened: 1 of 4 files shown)


evantahler:
closes https://github.com/actionhero/actionhero/issues/1408https://github.com/actionhero/actionhero/issues/1408

PR #1440, opened 2020-04-23
Add VSCode debug task for jest

evantahler:
We can use the integrated VSCode debugger to run Jest tests as well!  The normal breakpoints and debug console work as expected.

Downsides:
- If you are paused at a break point, that probably means your test will fail because it took too long (default 5 seconds)
- This VSCode task by default will run _every_ test in your project.  You will need to modify the `runtimeArgs` to pass in a single test file. 

Screen Shot 2020-04-22 at 6 27 19 PM


evantahler:
CC @allyusd and @gcoonrod!

PR #1441, opened 2020-04-25
Replace circleCI with Github actions

evantahler:
A test to see what changes would be needed to run Actionhero's tests wihin Github actions vs Circle CI


evantahler:
For a little while, we'll be testing with both CircleCI and Github Actions, to make sure that there are no surprises.

Issue #1472, opened 2020-05-23
The demo index.html that ships with Actionhero should not use ActionheroWebsocketClient


evantahler:
We ship this [example page](https://github.com/actionhero/actionhero/blob/master/public/index.html) with all new Actionhero projects created with `actionhero generate`.  It currently uses `ActionheroWebsocketClient` to make API requests.  However, if you disable the webscoket server, `ActionheroWebsocketClient` will not be generated, and there will be error on the page.

This page should be changed to use `fetch`.

PR #1473, opened 2020-05-23
replace ActionheroWebsocketClient with fetch in example index.html

evantahler:
This requires a modern-ish browser that actually has `fetch`... but that's the majority of browsers now.  closes https://github.com/actionhero/actionhero/issues/1472


evantahler:
closes https://github.com/actionhero/actionhero/issues/1472

Issue #1470, opened 2020-05-20
Replace`showDocumenatation` with Swagger/OpenAPI


evantahler:
For _years_, Actionhero has shipped with a `showDocumentation` action that provided a way to describe the actions running on your server.  The action was used to self-document the abilities of your server.    However, the format of that action is arbitrary... and not helpful in a larger ecosystem.  

Switching the format of this action to [swagger/OpenAPI](https://swagger.io/specification/v2/) will make the Action much more useful

--- 

As a proof of concept, here's a possible version of this action assuming action names follow the convention of `topic:name` and routes look like `/api/v{versionNumber}/action`

```ts
import { Action, config, api } from "actionhero";
import fs from "fs";
import path from "path";

const SWAGGER_VERSION = "2.0";
const API_VERSION = 1;
const parentPackageJSON = JSON.parse(fs.readFileSync(path.join(__dirname, '..', '..', 'package.json')));

const responses = {
  200: {
    description: "successful operation",
  },
  400: {
    description: "Invalid input",
  },
  404: {
    description: "Not Fount",
  },
  422: {
    description: "Missing or invalid params",
  },
  500: {
    description: "Server error",
  },
};

export class Swagger extends Action {
  constructor() {
    super();
    this.name = "swagger";
    this.description = "return API documentation";
    this.outputExample = {};
  }

  getLatestAction(route) {
    let matchedAction;
    Object.keys(api.actions.actions).forEach((actionName) => {
      Object.keys(api.actions.actions[actionName]).forEach((version) => {
        const action = api.actions.actions[actionName][version];
        if (action.name === route.action) {
          matchedAction = action;
        }
      });
    });

    return matchedAction;
  }

  buildSwaggerPaths() {
    const swaggerPaths = {};
    const tags = [];

    Object.keys(api.routes.routes).map((method) => {
      api.routes.routes[method].map((route) => {
        const action = this.getLatestAction(route);

        const tag = action.name.split(":")[0];
        const formattedPath = route.path
          .replace("/v:apiVersion", "")
          .replace(/\/:(\w*)/, "/{$1}");

        swaggerPaths[formattedPath] = swaggerPaths[formattedPath] || {};
        swaggerPaths[formattedPath][method] = {
          tags: [tag],
          summary: action.description,
          // description: action.description, // this is redundant
          consumes: ["application/json"],
          produces: ["application/json"],
          parameters: Object.keys(action.inputs)
            .sort()
            .map((inputName) => {
              return {
[...]

PR #1474, opened 2020-05-24
Swagger API Documentation

evantahler:
For years, Actionhero has shipped with a `showDocumentation` action that provided a way to describe the actions running on your server. The action was used to self-document the abilities of your server. However, the format of that action is arbitrary... and not helpful in a larger ecosystem.

Switching the format of this action to swagger/OpenAPI will make the Action much more useful.

* Switch the `showDocumentation` action with `swagger` action. Remove all old references to `showDocumentation`
* Include a new `swagger.html` and `swagger` action in generated projects which can consume and demo the API
* Remove `initializer/documentation`, as we don't need this documentation internally without the old showDocumentation action. 

Screen Shot 2020-05-23 at 5 15 28 PM

The breaking change is only around the `showDocumentation` and `documentation` initializer, which appears to not be used very much.

Closes https://github.com/actionhero/actionhero/issues/1470


src/actions/showDocumentation.ts
```diff
@@ -1,87 +0,0 @@
-import { api, Action } from "./../index";
```
gcoonrod, line 1: Any reason to leave this action around for backwards compatibility?

src/actions/swagger.ts
```diff
@@ -0,0 +1,147 @@
...
+          return;
+        }
+
+        swaggerPaths[formattedPath] = swaggerPaths[formattedPath] || {};
+        swaggerPaths[formattedPath][method] = {
+          tags: [tag],
+          summary: action.description,
+          // description: action.description, // this is redundant
+          consumes: ["application/json"],
+          produces: ["application/json"],
```
gcoonrod, line 78: Do our actions _always_ consume and produce `application/json`?

src/actions/swagger.ts
```diff
@@ -0,0 +1,147 @@
...
+                type: "string", // not really true, but helps the swagger validator
+                required:
+                  action.inputs[inputName].required ||
+                  route.path.includes(`:${inputName}`)
+                    ? true
+                    : false,
+                default:
+                  action.inputs[inputName].default !== null &&
+                  action.inputs[inputName].default !== undefined
+                    ? typeof action.inputs[inputName].default === "object"
```
gcoonrod, line 94: Typescript doesn't give us something better than `typeof of === "object"`?

evantahler:
Great questions @gcoonrod, and thanks for the review!  

I think that this action is a lot like `status` - you are probably going to want to customize it for your application.  Does you API only speak `application/json`?   Do you want to add tags?  Do you want to hide duplicate routes?  No idea!  I think I was trying to start with a 'common enough' boilerplate that would get folks started.  That's why I like having this logic contained entirely in an action (userland) vs in an initializer deep within Actionhero that would be harder to customize. 

As far as backwards-compatibility goes, I got the sense that no one was really using the `documentation` actions/initializers... but I'll ask again in Slack!

src/actions/swagger.ts
evantahler: not that I've found yet :(

src/actions/swagger.ts
```diff
@@ -0,0 +1,147 @@
...
+      //   return { name: tag, description: `topic: ${tag}` };
+      // }),
+      schemes: config.servers.web.allowedRequestHosts[0]
+        ? ["https", "http"]
+        : ["http"],
+      paths: swaggerPaths,
+
+      securityDefinitions: {
+        // TODO (custom)?
+      },
```
evantahler, line 140: Here's an example https://github.com/grouparoo/grouparoo/blob/master/core/api/src/actions/swagger.ts#L134-L145

PR #1494, opened 2020-06-08
Add action response to logger and user can disable and filter private…

ShunPin:
I add action response object to logger.
User can disable this by set enableResponseLogging = false in config/api.ts.
And when the response object contain private information, it can be filter by set filteredResponse to whatever you want.

### DIFF
--- a/__tests__/utils/utils.ts
+++ b/__tests__/utils/utils.ts
@@ -335,4 +335,86 @@ describe("Utils", () => {
       expect(filteredParams.o1.o2.o2p2).toEqual(testInput.o1.o2.o2p2);
     });
   });
+
+  describe("utils.filterResponseForLogging", () => {
+    beforeEach(() => {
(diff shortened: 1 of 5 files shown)


evantahler:
This is great work @ShunPin! Thank you for doing this!   Good testing too!

My only comment is that I think that by default, logging the response should be disabled.  When comparing similar features with other frameworks, and nginx & rails - they all do not log the response payload.  I believe that we should follow the same default pattern to, by default, protect user data, which is very likely to be logged in the response payload.

evantahler:
Also, I just updated https://github.com/actionhero/actionhero/commit/0c250babec50289bda6528ee18c841c4cea9b108 so the test suite should no run on all Pull Requests

ShunPin:
I forgot to remove testing log.
so a new PR just now.

evantahler:
Awesome work! Thank you!

PR #1512, opened 2020-07-01
`config.tasks.queues` can be a static array or an async function that returns list of queues

evantahler:
This PR allows `api.tasks.queues` to be either a static array or an async function that returns a list of queues.  This method will be called once when the server starts.  In this way, you can provide custom logic to build up your list of queues for this server.

```ts
// config/tasks.ts

export const DEFAULT = {
  tasks: (config) => {
    return {
      _toExpand: false,

      // Should this node run a scheduler to promote delayed tasks?
      scheduler: false,

      // what queues should the taskProcessors work?
      queues: ["*"],
      // Or, rather than providing a static list of `queues`, you can define a method that returns the list of queues.
      queues: async () => { return ["queueA", "queueB"]; },

// ...

```

### DIFF
--- a/__tests__/core/tasks/customQueueFunction.ts
+++ b/__tests__/core/tasks/customQueueFunction.ts
@@ -1,0 +1,67 @@
+import { api, Process, utils, config } from "./../../../src/index";
+
+const actionhero = new Process();
+let taskOutput = [];
+const queue = "testQueue";
+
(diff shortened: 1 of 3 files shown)


evantahler:
docs updated via https://github.com/actionhero/www.actionherojs.com/commit/69577ced0eec7d037b7bd0fdb4f509996668ae74

PR #1514, opened 2020-07-02
New Actionhero Projects should generate a swagger.html file

evantahler:
Co-Authored with @gcoonrod @stylder!

Correct a bug that forget to give new projects the actual swagger html file :(


evantahler:
@crrobinson14 maybe this was why you were confused :(

PR #1519, opened 2020-07-07
Fix TaskData type

witem:
`TaskData` defined as object with `params`: https://github.com/actionhero/actionhero/blob/ddc4821774f75c9fe0835143e61b99a3e7dab548/src/classes/task.ts#L92-L96

But when some task enqueues it not adds `params` cover, simply send `inputs` as arg: https://github.com/actionhero/actionhero/blob/ddc4821774f75c9fe0835143e61b99a3e7dab548/src/modules/task.ts#L56-L62


evantahler:
Thank you!  

How would you feel about also using the `TaskData` type in `modules/task` (replacing `{ [key: string]: any }`) to indicate that the same data will be returned back to the task when run?

ie: 
https://github.com/actionhero/actionhero/blob/ddc4821774f75c9fe0835143e61b99a3e7dab548/src/modules/task.ts#L58
https://github.com/actionhero/actionhero/blob/ddc4821774f75c9fe0835143e61b99a3e7dab548/src/modules/task.ts#L77
https://github.com/actionhero/actionhero/blob/ddc4821774f75c9fe0835143e61b99a3e7dab548/src/modules/task.ts#L97
https://github.com/actionhero/actionhero/blob/ddc4821774f75c9fe0835143e61b99a3e7dab548/src/modules/task.ts#L117
https://github.com/actionhero/actionhero/blob/ddc4821774f75c9fe0835143e61b99a3e7dab548/src/modules/task.ts#L135
https://github.com/actionhero/actionhero/blob/ddc4821774f75c9fe0835143e61b99a3e7dab548/src/modules/task.ts#L152

witem:
@evantahler I think its good idea. I change it.
But maybe also need change argument name from `inputs` to `taskData`not to be confused with `inputs` in task  describe

evantahler:
That works for me! Feel free to change the name if you want!

witem:
I change arguments naming, check please. Thanks!

evantahler:
Ah! I should have been more clear, I'm sorry - Since we use the word `inputs` already in our examples and docs, I think we should keep that word.  I would suggest changing `TaskData` to `TaskInputs` in the exported interface.

witem:
Ok, I update

evantahler:
Awesome! Thanks!

PR #1520, opened 2020-07-08
feat: action validate inputs has reserved param

allyusd:
issue:
My colleague meets the issue, he adds an "action" key in an action input...
but "action" is a keyword, apiVersion also.
can we have a checker to avoid it? even in build time or runtime.
[Example](https://github.com/allyusd/actionhero-action-inputs-using-reserved-params/blob/master/src/actions/usingReservedParams.ts)

problem:
I was tried to add `class/action.ts` validate using `api.params.globalSafeParams` of `initializers/params.ts`

But `action.validate` called by `initializers/action.ts` and its load priority is 410 and `params.ts` is 420.
So `api.parmas` is undefined when `action.validate` is call.
I was tried to change load priority, but I found `params.ts` need `api.actions` two... It's a deadlock, have any idea?

### DIFF
--- a/src/classes/action.ts
+++ b/src/classes/action.ts
@@ -94,5 +94,13 @@ export abstract class Action {
         `action \`${this.name}\` is a reserved verb for connections. choose a new name`
       );
     }
+
+    Object.keys(this.inputs).forEach((input) => {
+      if (api.params.globalSafeParams.indexOf(input) >= 0) {
+        throw new Error(
+          `action \`${this.name}\` input \`${input}\` is a reserved param. choose a new input`
+        );
+      }
+    });
   }
 }


evantahler:
Great start!
For this PR to be accepted:
* You'll need to add a test that demonstrates the new behavior you've added (try to create an action that use a reserved input, call `validate` and see that it fails)
* You'll need to get the test suite that exists passing.  Have you run `npm test` locally and seen the error?

src/classes/action.ts
```diff
@@ -94,5 +94,13 @@ export abstract class Action {
         `action \`${this.name}\` is a reserved verb for connections. choose a new name`
       );
     }
+
+    Object.keys(this.inputs).forEach((input) => {
+      if (api.params.globalSafeParams.indexOf(input) >= 0) {
+        throw new Error(
+          `action \`${this.name}\` input \`${input}\` is a reserved param. choose a new input`
```
evantahler, line 101: This might be an english language thing, but tt's common for error messages to be one sentence.  I'd remove "choose a new input".

evantahler:
I've taken your orignal work and made a new PR over at https://github.com/actionhero/actionhero/pull/1521

PR #1505, opened 2020-06-25
Action and Task errors log on only one line

evantahler:
closes https://github.com/actionhero/actionhero/issues/1504

Actions and tasks now log to one line in the file and other transports.  The stack trace contains `\n` so it still is legible pretty in the console. 

**Action**
```bash
# Console
2020-06-25T00:48:24.045Z - error: [ action @ web ] to=127.0.0.1 action=broken params={"action":"broken","apiVersion":1} duration=1 error=TypeError: Cannot read property 'missing' of undefined stacktrace=TypeError: Cannot read property 'missing' of undefined
    at ValidationTest.run (/Users/evan/workspace/actionhero/actionhero/src/actions/broken.ts:14:27)
    at ActionProcessor.runAction (/Users/evan/workspace/actionhero/actionhero/src/classes/actionProcessor.ts:378:35)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)

# File
{"to":"127.0.0.1","action":"broken","params":"{\"action\":\"broken\",\"apiVersion\":1}","duration":1,"error":"TypeError: Cannot read property 'missing' of undefined","stacktrace":"TypeError: Cannot read property 'missing' of undefined\n    at ValidationTest.run (/Users/evan/workspace/actionhero/actionhero/src/actions/broken.ts:14:27)\n    at ActionProcessor.runAction (/Users/evan/workspace/actionhero/actionhero/src/classes/actionProcessor.ts:378:35)\n    at processTicksAndRejections (internal/process/task_queues.js:97:5)\n    at WebServer.processAction (/Users/evan/workspace/actionhero/actionhero/src/classes/server.ts:229:18)","level":"error","message":"[ action @ web ]","timestamp":"2020-06-25T00:48:24.045Z"}

```

**Task**

```bash
# Console
2020-06-25T00:38:03.840Z - error: error from task: task:busted on queue default (worker #1) arguments=[object Object] stacktrace=Error: NOPE
    at MyTask.run (/Users/evan/workspace/actionhero/actionhero/src/tasks/busted.ts:14:11)
    at Worker.perform (/Users/evan/workspace/actionhero/actionhero/src/initializers/tasks.ts:123:39)
    at Worker.perform (/Users/evan/workspace/actionhero/actionhero/node_modules/node-resque/dist/core/worker.js:165:41)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)

# File
[...]

PR #1558, opened 2020-08-14
Brings back api.exceptionHandlers.action for logging action errors with third-party error reporters

evantahler:
Returns `api.exceptionHandlers.action` so that errors from actions can be passed to third-party error reporters.  IE:

```ts
api.exceptionHandlers.reporters.push((error: Error) => {
  Sentry.captureException(error);
});
```

The `actionProcessor` still handles logging the action.

Related to https://github.com/actionhero/ah-sentry-plugin


evantahler:
This was erroneously removed in https://github.com/actionhero/actionhero/pull/1505, as it does more than just handle writing to the console.

PR #1562, opened 2020-08-19
Browser tests with Puppeteer rather than Selenium

evantahler:
In theory, Puppeteer is will be easier for folks to use.  It works cross-platform, knows how to install it's own version of Chrome and doesn't have an external dependency like `chrome-webdriver`.   

If only I can get it to run on CI...

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -95,7 +95,8 @@ jobs:
             node_modules
             dist
           key: ${{ runner.os }}-cache-${{ github.sha }}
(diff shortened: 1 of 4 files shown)


gcoonrod:
What is the reasoning behind this change?

evantahler:
Updated the description!

evantahler:
Ideas from Clay (from slack) https://gist.github.com/chimmelb/6342504893b2b9fce0f4a8efd096ae60

evantahler:
Thanks to @allyusd and #1654 the tests are passing!

gcoonrod (approved):

PR #1570, opened 2020-08-26
Adds an option to override default error log formatting

greenzelenetskyi:
Move action error logline formatting logic from actionProcessor to error serializers to make it possible to override default action error logging, as previously discussed in https://github.com/actionhero/actionhero/discussions/1561


greenzelenetskyi:
Hi, I'm not sure, what could've caused the failing test after the merge. I'd appreciate any help.

evantahler:
Looks like it was just a timeout in the Browser integration test:

```
    Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Error: Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.
```

I'll run them again! In the future, if you want to run it again yourself, just push another empty commit `git commit --allow-empty -m "Trigger Build"`

witem:
> This looks like a great addition!
> 
> I'd suggest adding a little more safety around your call to `config.errors.serializers.actionProcessor`. As written, this would be a breaking change to Actionhero, as it won't work without adding the new method to config/errors. Is there a (possibly simplified) default behavior you can add if `config.errors.serializers.actionProcessor` doesn't exist?

But it also breaks old log format. Maybe need to move default serializer to `classes/actionProcessor.ts`. And add possibility to redefine it over `config.errors.serializers.actionProcessor` if it exists.

greenzelenetskyi:
Great points. Thanks. I've moved the default logic back to actionProcessor and you use it as a fallback to the serializer. What do you think?

src/config/errors.ts
```diff
@@ -31,6 +31,7 @@ export const DEFAULT = {
             }
           },
         },
+        actionProcessor: null,
```
evantahler, line 34: Add a note here with the method signature at least and where to look for how to customize. ie `See ActionProcessor#applyDefaultErrorLogLineFormat` to see an example of how to customize.

evantahler:
Awesome work!

PR #1580, opened 2020-09-03
Update Dockerfile to use Node alpine image instead of base alpine

gcoonrod:
The objective here is to reduce the amount of building that needs to happen when using Docker.

@evantahler how do I get rid of all these merge commits?


evantahler:
That's a lot of merging :D

docker-compose.yml
```diff
@@ -22,18 +22,33 @@ services:
     networks:
       - actionhero_frontend
       - actionhero_backend
+    volumes: 
+      - /actionhero/node_modules
+      - /actionhero/dist
```
evantahler, line 27: What do volumes do?  I think that means they will persist on (each? one?) node in the docker cluster?

docker-compose.yml
```diff
@@ -22,18 +22,33 @@ services:
...
   #
   # actionhero:
   #   build: ./
+  #   image: actionhero
   #   ports:
-  #     - 80:8080
+  #     - 8080
   #   environment:
-  #     TCP_PORTS: 8080
+  #     SERVICE_PORTS: 8080
```
evantahler, line 37: Did HAPROXY change what it looks for?

Dockerfile
```diff
@@ -1,12 +1,11 @@
-FROM alpine:latest
-MAINTAINER admin@actionherojs.com
+FROM node:alpine
```
evantahler, line 1: neat!

docker-compose.yml
gcoonrod: My understanding is that these volumes will ensure that any architecture differences that would impact module builds (i.e. node-gyp) between the host system and the container are eliminated by preventing these mounted directories from being shared with the host system. However, I'm no docker expert and these may not be strictly necessary.

docker-compose.yml
gcoonrod: I wasn't sure why the commented out build section differed from the uncommented section that pulls from docker hub.

gcoonrod:
Closing in favor of #1583

PR #1604, opened 2020-10-01
Export the winston loggers as "loggers" from Actionhero

evantahler:
You can now `import {loggers} from 'actionhero'` to access the winston loggers you created in `config/loggers`.


evantahler:
Docs updated in https://github.com/actionhero/www.actionherojs.com/pull/189

PR #1605, opened 2020-10-02
Support retryStuckJobs option for Resque Scheduler

evantahler:
As of https://github.com/actionhero/node-resque/pull/450 (released in `node-resque` v7.2.0), the Scheduler can be set to automatically retry jobs/tasks which have been failed for being 'stuck' (resulting in errors like "Worker Timeout (killed manually)").

The Actionhero config option, `config.tasks.retryStuckJobs` (boolean) will be passed to the `node-resque` scheduler to enable this new feature.  By default, the option is disabled to maintain the existing behavior of 'suck' tasks not being retried automatically.


evantahler:
This Pull Request also updates all packages to their newest versions

PR #1607, opened 2020-10-03
Add automaticRoutes, Remove query and simple routing

evantahler:
Per the discussion in https://github.com/actionhero/actionhero/discussions/1579, Actionhero is removing the options `config.servers.web.simpleRouting` and `config.servers.web.queryRouting`.  These options were confusing, and generally conflicted with the `config/routes.ts` file.  We want to be clear to new users that the "right" way to define the routes for your Actionhero APIs is in `/config/routes.ts`.

That said, there are times when you may need to quickly build routes automatically from your actions, like when testing or deploying this core Actionhero project directly 😁.  For those cases, this PR adds a new option, `config.servers.web.automaticRoutes`.  This new option is similar-ish to the old `automaticRoutes` option, but differs in a few key ways:
* disabled by default.  New users will not get confused about why actions are showing up automatically
* can support some/many HTTP verbs.  The verb(s) you want are now declarative, ie:
  * `config.servers.web.automaticRoutes = ['get']`
  * `config.servers.web.automaticRoutes = ['get', 'post', 'head']`

For newly generated Actionhero Projects, we will start with your `config/routes.ts` file including a `get` entry for the Status, Swagger, and CreateChatRoom actions, so the examples continue to work

### DIFF
--- a/__tests__/integration/sharedFingerprint.ts
+++ b/__tests__/integration/sharedFingerprint.ts
@@ -31,6 +31,7 @@ const connectClient = async (query = ""): Promise<any> => {
 
 describe("Integration: Web Server + Websocket Socket shared fingerprint", () => {
   beforeAll(async () => {
+    process.env.AUTOMATIC_ROUTES = "get";
     await actionhero.start();
     await api.redis.clients.client.flushdb();
     url = "http://localhost:" + config.servers.web.port;
--- a/__tests__/servers/web/jsonp.ts
+++ b/__tests__/servers/web/jsonp.ts
@@ -17,8 +17,6 @@ jest.mock("./../../../src/config/servers/web.ts", () => ({
           rootEndpointType: "file",
           port: 18080 + parseInt(process.env.JEST_WORKER_ID || "0"),
           matchExtensionMime: true,
-          simpleRouting: true,
-          queryRouting: true,
           metadataOptions: {
             serverInformation: true,
             requesterInformation: false,
--- a/__tests__/servers/web/jsonp.ts
+++ b/__tests__/servers/web/jsonp.ts
@@ -11,7 +11,7 @@ jest.mock("./../../../src/config/servers/web.ts", () => ({
       web: () => {
         return {
           enabled: true,
-          automaticRoutes: "get,post",
+          automaticRoutes: ["get", "post"],
           secure: false,
           urlPathForActions: "api",
           urlPathForFiles: "public",
--- a/__tests__/servers/web/jsonp.ts
+++ b/__tests__/servers/web/jsonp.ts
@@ -11,6 +11,7 @@ jest.mock("./../../../src/config/servers/web.ts", () => ({
       web: () => {
         return {
           enabled: true,
+          automaticRoutes: "get,post",
           secure: false,
           urlPathForActions: "api",
           urlPathForFiles: "public",
--- a/src/config/servers/web.ts
+++ b/src/config/servers/web.ts
@@ -39,8 +39,12 @@ export const DEFAULT = {
         //  Visitors can always visit /api and /public as normal
         rootEndpointType: "file",
         // In addition to what's defined in config/routes.ts, should we make a route for every action?  Useful for debugging or simple APIs.
-        // automaticRoutes should a be comma-separated list of HTTP verbs, ie: null (default), 'get', 'post', 'get,put', 'get,post,put', etc.
-        automaticRoutes: process.env.AUTOMATIC_ROUTES,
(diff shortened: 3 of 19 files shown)


evantahler:
@gcoonrod we had discussed also updating the `actionhero generate action` script to automatically add an entry to `config/routes.ts` when used... but I ran into a few hiccups:

* How do we know you want to generate a `web` action?  It could be for `websocket` or a custom server.
* How do we know which verb you want the action to respond to?
* Do you want any of the required params to be in the route (ie: `/:apiVersion/user/:userId`)?

To answer these questions we would need to add a lot of CLI args/flags... which seems to defeat the purpose of "making it easy for new users"... so I left it off for now.

__tests__/servers/web/jsonp.ts
```diff
@@ -11,14 +11,13 @@ jest.mock("./../../../src/config/servers/web.ts", () => ({
       web: () => {
         return {
           enabled: true,
+          automaticRoutes: "get,post",
```
evantahler, line 14: Note how you can now opt-into the automaticRoutes you need per test suite.

evantahler:
Question: Should the config expect an array of strings, and do the parsing of the ENV in the config file?

__tests__/integration/sharedFingerprint.ts
```diff
@@ -31,6 +31,7 @@ const connectClient = async (query = ""): Promise<any> => {
 
 describe("Integration: Web Server + Websocket Socket shared fingerprint", () => {
   beforeAll(async () => {
+    process.env.AUTOMATIC_ROUTES = "get";
```
gcoonrod, line 34: We'll need to do this in every test now? If a developer forgets to do this will they get clear errors about what they're missing?

src/config/servers/web.ts
```diff
@@ -38,10 +38,13 @@ export const DEFAULT = {
...
         //  Visitors can always visit /api and /public as normal
         rootEndpointType: "file",
-        // simple routing also adds an 'all' route which matches /api/:action for all actions
-        simpleRouting: true,
-        // queryRouting allows an action to be defined via a URL param, ie: /api?action=:action
-        queryRouting: true,
+        // In addition to what's defined in config/routes.ts, should we make a route for every action?  Useful for debugging or simple APIs.
+        // automaticRoutes should an array of strings - HTTP verbs, ie: [] (default), ['get'], ['post'], ['get','put'], ['get','post','put'], etc.
+        automaticRoutes: process.env.AUTOMATIC_ROUTES
+          ? process.env.AUTOMATIC_ROUTES.split(",")
```
gcoonrod, line 44: Does it make sense to have all of this logic in a config file? Perhaps externalize it to a utility function.

src/initializers/routes.ts
```diff
@@ -206,22 +210,26 @@ export class Routes extends Initializer {
...
-          "debug"
-        );
 
-        log("routes:", "debug", api.routes.routes);
+          log(
+            `creating routes automatically for all actions to ${verb} HTTP verb`
+          );
+
+          for (const action in api.actions.actions) {
+            route.registerRoute(verb, "/" + action, action, null);
```
gcoonrod, line 227: Should we add a debug level log here listing each route as it is registered? Or does `registerRoute` do that for us?

__tests__/integration/sharedFingerprint.ts
evantahler: Good question!

I think we are ok:

1. no matter what, `specHelper.runAction()` will always work with whatever actions you have.  
2. this test makes /real/ HTTP requests, so we need the actions to be in the routing table.  However, we didn't add them explicitly because 
3. I don't want to have a different `routes.ts` in actionhero "core" from what gets generated

src/config/servers/web.ts
evantahler: I'm on the fence about this one.  How are folks going to use this?  
* Is the list static - if so, then we don't need any of this and just an array in the file is OK
* Is the list dynamic based on env?  Then we still don't need anything special outside of our normal per-env exports
* Is the list dynamic based on some developer setting?  Then me might need to parse the ENV

My main goal was to keep it only in "user" land so it could be modified as needed - sort of how we parse database credentials for ah-sequelize https://github.com/actionhero/ah-sequelize-plugin/blob/master/src/config/sequelize.ts

src/initializers/routes.ts
evantahler: We could!  At the very lease I added this per HTTP verb:

```
log(`creating routes automatically for all actions to ${verb} HTTP verb`);
```

PR #1610, opened 2020-10-06
Actions can return responses to determine response types

evantahler:
The preferred way to describe an Action's run method is now by returning an object which you want to send to your consumers.  By doing this, you are building up a Typescript type which can then be used in your API to type-check API responses! 

For example:

```ts
// src/actions/randomNumber.ts
import { Action } from "actionhero";

export class RandomNumber extends Action {
  constructor() {
    super();
    this.name = "randomNumber";
    this.description = "I am an API method which will generate a random number";
    this.outputExample = { randomNumber: 0.123 };
  }

  async run({ connection }) {
    const randomNumber = Math.random();
    const stringRandomNumber: string = connection.localize([
      "Your random number is {{randomNumber}}",
      { randomNumber },
    ]);

    return { randomNumber, stringRandomNumber };
  }
}
```

Now, you can load in the action and inspect the run method's return type:

```ts
const { RandomNumber } = await import("../../src/actions/randomNumber");
type ResponseType = UnwrapPromise<typeof RandomNumber.prototype.run>;

// now that we know the types, we can enforce that new objects match the type
// this is OK!
const responsePayload: ResponseType = {
  randomNumber: 1,
  stringRandomNumber: "some string",
};

// this fails compilation (missing stringRandomNumber):
const responsePayload: ResponseType = {
  randomNumber: 1,
};
```

The type information is also available to your IDE

Screen Shot 2020-10-06 at 4 33 30 PM

Setting `data.response` in the action is still possible, but now it will be discouraged by the class Action, which expects the run method to return an object or null. 

Actionhero now includes the helper types `<UnwrapPromise>` and `<AssertEqualType>` for these types of use cases.

### DIFF
--- a/__tests__/actions/swagger.ts
+++ b/__tests__/actions/swagger.ts
@@ -17,7 +17,7 @@ describe("Action", () => {
       const { paths, basePath, host } = await specHelper.runAction("swagger");
       expect(basePath).toBe("/api/");
       expect(host).toMatch(/localhost/);
-      expect(Object.keys(paths).length).toEqual(7); // 7 actions
+      expect(Object.keys(paths).length).toEqual(8); // 8 actions
     });
   });
 });
--- a/__tests__/core/api.ts
+++ b/__tests__/core/api.ts
@@ -1,4 +1,12 @@
-import { config, api, Process, Action, specHelper } from "./../../src/index";
+import {
+  config,
+  api,
+  Process,
+  Action,
+  specHelper,
+  UnwrapPromise,
+  AssertEqualType,
+} from "./../../src/index";
 
 const actionhero = new Process();
 
@@ -63,8 +71,8 @@ describe("Core", () => {
             description: "I am a test",
             version: 1,
             outputExample: {},
-            run: async (data) => {
-              data.response.version = 1;
+            run: async () => {
+              return { version: 1 };
             },
           },
           //@ts-ignore
@@ -73,8 +81,8 @@ describe("Core", () => {
             description: "I am a test",
             version: 2,
             outputExample: {},
-            run: async (data) => {
-              data.response.version = 2;
+            run: async () => {
+              return { version: 2 };
             },
           },
           //@ts-ignore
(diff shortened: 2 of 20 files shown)


evantahler:
These responses will be merged with the responses of any pre or postProcessor middlewares

mfvargo:
I think it makes intuitive sense for the run return value to be the response.  For me this only would be useful in a middleware postprocess  function to have the typecheck.  I like the UnwrapPromis thing.  That's nice.

How would you do things where you send a file?
```
  async run(data) {
    const { params, connection } = data;
    const attachments = await connection.contest.dispute.getAttachments({
      where: { id: params.attachmentId },
    });
    if (!attachments[0]) {
      throw new Error("attachment does not belong to this dispute");
    }
    connection.sendFile(attachments[0].path);
    data.toRender = false;
  }
```
?

evantahler:
@mfvargo that's a good question!

I added an example and test via https://github.com/actionhero/actionhero/pull/1610/commits/a09aa545323bad5fdd5f66f9447ec2dbd08e0eb6 to ensure that this continues to work

mfvargo:
@evantahler Well, i guess it would depend on how the middleware might want to know things about the file being sent.  I'm trying to think of some example.  Maybe the middleware wants to run a virus check or something on the file?  So having a way to know that might be important.

I guess if connection.sendFile() returned some data the run function could just return that and that type would be shown to the midddleware.  

```
async run(data) {
...
  return connection.sendFile("pathtosomefile");
}
```

evantahler:
I don't think I'm following @mfvargo - are you asking for a new feature, for `connection.sendFile()` to change the TypeScript response type of the run method to be something like `{fileName: string}`?   If so, you would need to really do that in the run method ie: `return {filename: 'filepath'}`.  If you do that, then it will be available in action postProcessors on `data.response.filename`.

mfvargo:
@evantahler Yeah, I can see that.  Maybe a run function that used to set data.toRender = false could instead just return null (exit with out a return statement)?

evantahler:
@mfvargo can you open a new issue or discussion with an example?  I don't think I'm getting it...

PR #1611, opened 2020-10-07
Use scan to read collections of redis keys and provide getKeys as a generic way to search for redis keys

evantahler:
Previous to this PR, `cache.keys()` and `cache.locks()` were using the `keys` redis command.  With large datasets, this can be slow and block other connections from doing work.  The better way is to use the scan command repeatedly with a cursor to not block the server for too long.  

This PR adds `cache.getKeys(pattern:string)` which uses `scan`.  `cache.keys()` and `cache.locks()` have been refactored to use `cache.getKeys`.  

The `cache.clean` method has been updated to accept an optional `pattern` for the keys to remove, as well as use a redis `pipeline` to speed things up.  Rather than make many requests per key, it will make one to delete all the keys it found.

There's a new config option required, `config.redis.scanCount` to provide how many keys we should scan for in each request.   The default is 1000

### DIFF
--- a/src/modules/cache.ts
+++ b/src/modules/cache.ts
@@ -16,10 +16,11 @@ export namespace cache {
     retry?: boolean | number;
   }
 
-  export const redisPrefix = config.general.cachePrefix;
-  export const lockPrefix = config.general.lockPrefix;
-  export const lockDuration = config.general.lockDuration;
-  export const lockRetry = 100;
+  export const redisPrefix: string = config.general.cachePrefix;
+  export const lockPrefix: string = config.general.lockPrefix;
+  export const lockDuration: number = config.general.lockDuration;
+  export const scanCount: number = config.redis.scanCount || 1000;
+  export const lockRetry: number = 100;
 
   export function client() {
     if (config.redis.enabled && api.redis.clients && api.redis.clients.client) {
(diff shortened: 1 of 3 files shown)


src/modules/cache.ts
```diff
@@ -41,51 +42,78 @@ export namespace cache {
     lockNameOverride = name;
   }
 
+  /**
+   * A generic method to find all keys which match a pattern.
+   * @pattern likely has a * at the end of the string.  Other arguments are for recursion and not required.
+   */
+  export async function getKeys(
```
bleonard, line 49: To clarify current situation: when using this for something in the cache, you'd want to call `cache.getKeys(cache.redisPrefix + "myPrefix" + "*")`
It might expected that a caller wouldn't have to prepend the `cache.redisPrefix` and it's pre-scoped like that.

bleonard: maybe `cache.keys()` should be more like
```
function keys(optionalPrefix="") {
  const pattern = redisPrefix + optionalPrefix + "*";
  return getKeys(pattern);
}
```

evantahler: love it -> https://github.com/actionhero/actionhero/pull/1612

PR #1628, opened 2020-10-25
Actions can be called in-line

evantahler:
This PR introduces `action.run()` which allows Actions to be run arbitrarily from elsewhere in the codebase.   With this, you can have one Action call another, have a Task run Action, etc.   Middleware for the the Action will be run.  `action.run()` creates a new proxy connection with only the arguments you provide to the method.

```ts
import { action } from "actionhero"

const nameOfAction = 'myAction'
const actionVersion = 'v1' // or leave null to use the latest version
const params = {key: 'value'} // the params which would be parsed from the client
const connectionProperties = {} // special properties on the connection which may be expected by the action or middleware.  Perhaps "session.id" or "authenticated = true" depending on your middleware

const response = await action.run(nameOfAction, actionVersion, params, connectionProperties);
```

So, if you wanted an Action which combines the responses of 2 other Actions:

```ts
import { Action, action } from "actionhero";

export class RecursiveAction extends Action {
  constructor() {
    super();
    this.name = "recursiveAction";
    this.description = "I am an action that runs 2 other actions";
    this.outputExample = {};
  }

  async run() {
    const localResponse = { local: true };
    const firstActionResponse = await action.run("otherAction");
    const secondActionResponse = await action.run("anotherAction");
    return Object.assign(firstActionResponse, secondActionResponse, localResponse);
  }
}
```

Future work: 
* Type the responses of `action.run()` according the run method's return types.

### DIFF
--- a/__tests__/actions/recursiveAction.ts
+++ b/__tests__/actions/recursiveAction.ts
@@ -1,0 +1,22 @@
+import { Process, specHelper } from "./../../src/index";
+
+const actionhero = new Process();
+
+describe("Action", () => {
+  describe("recursiveAction", () => {
+    beforeAll(async () => {
+      await actionhero.start();
+    });
+
+    afterAll(async () => {
+      await actionhero.stop();
+    });
+
+    test("merges its own response with the randomNumber response", async () => {
+      const response = await specHelper.runAction("recursiveAction");
+      expect(response.local).toEqual(true);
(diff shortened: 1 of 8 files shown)


evantahler:
A number of use cases can be seen in the tests https://github.com/actionhero/actionhero/blob/5338b49b194b02bf005ae67cdd16860a55841d12/__tests__/modules/action.ts

PR #1643, opened 2020-11-08
Replace up to 5 params in routes for swagger

evantahler:
In routes that contain lots of params, ie `/api/:username/:post/:comment`, swagger doesn't properly replace params after the first one.


evantahler:
Merging... but noting that there are probable better ways to do this.

PR #1648, opened 2020-11-11
doc: chatRoom broadcast, improve doc header 

dmpvost:
Try to improve the explanation of parameters, but finally, I'm note sure that it's better, 

feel free to participate on this PR


evantahler:
👍

PR #1653, opened 2020-11-17
Actionhero can be run with ioredis-mock (Redis is always enabled)

evantahler:
This PR allows Actionhero to be run with the [`ioredis-mock`](https://github.com/stipsan/ioredis-mock) package.  In this way, we can allow users of Actionhero to use **every** features of the framework, including cache, chat, and tasks, without a Redis server running locally.   This is a breaking change as it removes the `enabled` option on `config.redis`, as Redis is now always enabled, but possibly mocked.

Of course, no data will be persisted or shared between nodes unless you are using a "real" Redis server.  The default configuration of Actionhero (when a new project is generated with `npx actionhero generate`) is to assume that a real Redis server can be reached at `localhost`.  However, we make it easy to change by providing a `ioredis-mock`-based configuration commented out in `config/redis.ts`.

Uses `ioredis-mock` may also be beneficial in your tests.

### DIFF
--- a/__tests__/integration/ioredis-mock.ts
+++ b/__tests__/integration/ioredis-mock.ts
@@ -1,0 +1,190 @@
+import * as MockIORedis from "ioredis-mock";
+import {
+  api,
+  cache,
+  chatRoom,
+  Process,
+  specHelper,
+  utils,
+  Task,
+  task,
+} from "./../../src/index";
+
+jest.mock("./../../src/config/redis.ts", () => ({
+  __esModule: true,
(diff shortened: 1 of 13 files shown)


evantahler:
Many years ago, we used to use https://github.com/hdachev/fakeredis for this purpose, but it hasn't been maintained.

gcoonrod (approved):
Hurray!

evantahler:
OK! Looks like we are OK with a breaking change then!

PR #1654, opened 2020-11-17
fixed ci fail for puppeteer

allyusd:
install puppeteer dependencies
launch puppeteer with no-sandbox

ref: puppeteer/puppeteer#3443 (comment)


evantahler:
Thank you!

Issue #1509, opened 2020-07-01
Wrong response logging with custom error handler


witem:
# Wrong response logging with custom error handler

If server use custom error formatter in logs we don't see correct response. Code for reproduce: https://github.com/witem/actionhero/commit/554153c07e8245b856848fee62b2536e97ec44fc .

When I run  `curl http://localhost:8080/api/testAppErr`, receive:

```
"error": {
    "code": "100500",
    "message": "some app error"
  },
```

But in console:

```
error: [ action @ web ] to=127.0.0.1 action=testAppErr params={"action":"testAppErr","apiVersion":1} duration=3 error=AppError: some app error response={"error":{"name":"AppError","appCode":"100500"}} name=AppError appCode=100500 stack=AppError: some app error
    at TestAppError.run (/home/ark/projects/git-projects/actionhero/src/actions/randomNumber.ts:43:17)
    at ActionProcessor.runAction (/home/ark/projects/git-projects/actionhero/src/classes/actionProcessor.ts:381:35)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at WebServer.processAction (/home/ark/projects/git-projects/actionhero/src/classes/server.ts:229:18) 

```

When I run  `curl http://localhost:8080/api/testErr`, receive:

```
"error": {
    "code": 500,
    "message": "Internal server error"
  }
```

But in console:

```
error: [ action @ web ] to=127.0.0.1 action=testAppErr params={"action":"testAppErr","apiVersion":1} duration=2 error=AppError: some app error response={"error":{"name":"AppError","code":"100500"}} name=AppErro
r code=100500 stack=AppError: some app error                                                                                                                                                                                                 
    at TestAppError.run (/home/ark/projects/git-projects/actionhero/src/actions/randomNumber.ts:43:17)                                                                                                                                       
    at ActionProcessor.runAction (/home/ark/projects/git-projects/actionhero/src/classes/actionProcessor.ts:381:35)                                                                                                                          
    at processTicksAndRejections (internal/process/task_queues.js:97:5)                                                                                                                                                                      
    at WebServer.processAction (/home/ark/projects/git-projects/actionhero/src/classes/server.ts:229:18)

```

- Actionhero Version: `23.0.0`
- Node.js Version: `12.18.0`
- Operating System: `Manjaro 5.7.0`

# Steps to reproduce your error

Example changes in: https://github.com/witem/actionhero/commit/554153c07e8245b856848fee62b2536e97ec44fc
[...]

PR #1661, opened 2020-11-25
Default action logging output to match action result returned to web client.

gcoonrod:
Hopefully, this addresses #1509. In addition, for actions handled by a web server, any thrown errors will result in the response HTTP status code of 500 (indicating a server error) instead of 400 (usually reserved for client errors).

In addition, this change adds an option to the web server config: `defaultErrorStatusCode` which is defaulted to 500. This option is only effective if the status code has not been set by the action.


evantahler:
Looks great!

evantahler (approved):

PR #1792, opened 2021-04-22
[cli] add support for custom placeholder and variadic options

pedroslopez:
When defining a CLI command, I wanted to be able to use a [Variadic option](https://github.com/tj/commander.js/#variadic-option), but did not find a way to do this through actionhero. Commander supports this by adding `...` to the placeholder so you'll get back an array of all the options the user specified when running the command.

This also adds the ability to customize the value placeholder to pluralize it in the help section or use a different word entirely.

Currently this is possible by adding the ellipsis to the key, but it has some consequences on the usage of the param later on.


evantahler:
Thanks @pedroslopez!

evantahler:
This kind of thing will be pretty hard to test... so I'm OK with not having a test on this for now

PR #1853, opened 2021-06-10
Better types for ExceptionReporter & use throughout

evantahler:
This PR refactors the Exception initializer and introduces some better types for a new `ExceptionReporter` type.  We then use `api.exceptionHandlers.report()` in the main process if something goes wrong rather than manually logging errors to the logger as we had been.

This PR's goal is to set up the `ExceptionReporter` for use elsewhere - eg perhaps you are using Sequelize which has [custom error types](https://sequelize.org/master/class/lib/errors/validation-error.js~ValidationError.html), and you want to make a new custom `ExceptionReporter` that gets these Sequelize errors and logs the SQL statement which caused the problem... hypothetically.

### DIFF
--- a/src/classes/exceptionReporter.ts
+++ b/src/classes/exceptionReporter.ts
@@ -1,0 +1,7 @@
+export type ExceptionReporter = (
+  error: Error,
+  type: string,
+  name: string,
+  objects?: any,
+  severity?: string
+) => void;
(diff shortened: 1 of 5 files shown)


evantahler:
This is likely a minor version change, `v26.1.0`, because these(internal) APIs did change slightly

PR #1870, opened 2021-06-29
Action Logging Improvements to aid `404` debugging

evantahler:
This PR improves a few things about Action logging:

1. [Enhancement] Action logs now include the `method` and `pathname` of the request, as seen by Actionhero, for web connections.

```
2021-06-30T00:05:53.156Z - info: [ action @ web ] to=127.0.0.1 action=createChatRoom params={"action":"createChatRoom","apiVersion":1,"name":"some room"} duration=1 method=GET pathname=/api/createChatRoom

2021-06-30T00:05:42.163Z - error: [ action @ web ] to=127.0.0.1 action=createChatRoom params={"action":"createChatRoom","apiVersion":1,"name":"foo"} duration=1 error=Error: room exists method=GET pathname=/api/createChatRoom stack=Error: room exists
    at Object.add (/Users/evan/workspace/actionhero/actionhero/src/modules/chatRoom.ts:111:13)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at CreateChatRoom.run (/Users/evan/workspace/actionhero/actionhero/src/actions/createChatRoom.ts:16:25)
    at ActionProcessor.runAction (/Users/evan/workspace/actionhero/actionhero/src/classes/actionProcessor.ts:427:32)
    at WebServer.processAction (/Users/evan/workspace/actionhero/actionhero/src/classes/server.ts:229:18)
```

2. [Breaking-ish] Errors of the type `unknown_action`, which normally produce a localized error of `"unknown action or invalid apiVersion"`, are no longer sent to the error reporter by default.  This is configured by the new `config.errors.reportUnknownActions` config setting.  When disabled (`config.errors.reportUnknownActions=false`), this means that the error will still be logged just as it was before, but if you are using an additional error reporter, like New Relic or Sentry, it will not get these error types.  
 
```
2021-06-30T00:07:18.938Z - error: [ action @ web ] to=127.0.0.1 params={"action":null,"apiVersion":null} duration=0 error=Error: unknown action or invalid apiVersion method=GET pathname=/api/not/a/real/thing stack=Error: unknown action or invalid apiVersion
    at ActionProcessor.completeAction (/Users/evan/workspace/actionhero/actionhero/src/classes/actionProcessor.ts:107:44)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

With these 2 improvements, it should be easier to debug 404s!

### DIFF
--- a/src/classes/actionProcessor.ts
+++ b/src/classes/actionProcessor.ts
@@ -137,9 +137,9 @@ export class ActionProcessor<ActionClass extends Action> {
       action: this.action,
       params: JSON.stringify(filteredParams),
       duration: this.duration,
-      error: "",
       method: type === "web" ? rawConnection.method : undefined,
       pathname: type === "web" ? rawConnection.parsedURL.pathname : undefined,
+      error: "",
       response: undefined,
     };
 
--- a/src/classes/actionProcessor.ts
+++ b/src/classes/actionProcessor.ts
@@ -160,7 +160,10 @@ export class ActionProcessor<ActionClass extends Action> {
 
     log(`[ action @ ${this.connection.type} ]`, logLevel, logLine);
 
-    if (error && status !== "unknown_action") {
+    if (
+      error &&
+      (status !== "unknown_action" || config.errors.reportUnknownActions)
+    ) {
       api.exceptionHandlers.action(error, logLine);
     }
   }
--- a/src/classes/actionProcessor.ts
+++ b/src/classes/actionProcessor.ts
@@ -5,6 +5,17 @@ import { log } from "../modules/log";
 import { utils } from "../modules/utils";
 import * as dotProp from "dot-prop";
 import { api } from "../index";
+import { stat } from "fs";
+
+export type ErrorStatusMessage =
+  | "complete"
+  | "generic_error"
+  | "server_shutting_down"
+  | "too_many_requests"
+  | "unknown_action"
+  | "unsupported_server_type"
+  | "missing_params"
+  | "validator_errors";
 
 export class ActionProcessor<ActionClass extends Action> {
   connection: Connection;
@@ -27,7 +38,7 @@ export class ActionProcessor<ActionClass extends Action> {
     [key: string]: any;
   };
   duration: number;
-  actionStatus: string | Error;
+  actionStatus: ErrorStatusMessage;
 
   // allow for setting of any value via middleware
   session: any;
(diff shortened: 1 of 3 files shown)


evantahler:
I don't think that this should be a major version breaking change.  While this PR does change behavior, I think that the expected behavior would have been that 404s don't get set to your APM tools... so I view this more as a bugfix.

That said, we can bump the minor version to indicate that there is /something/ that changed.  Thoughts?

src/classes/actionProcessor.ts
```diff
@@ -5,6 +5,17 @@ import { log } from "../modules/log";
 import { utils } from "../modules/utils";
 import * as dotProp from "dot-prop";
 import { api } from "../index";
+import { stat } from "fs";
+
+export type ErrorStatusMessage =
```
gcoonrod, line 10: General question: any reason not to use an Enum here?

src/classes/actionProcessor.ts
```diff
@@ -90,15 +101,14 @@ export class ActionProcessor<ActionClass extends Action> {
...
     }
 
-    if (error && typeof error === "string") {
-      error = new Error(error);
-    }
+    if (typeof error === "string") error = new Error(error);
 
     if (error && (typeof this.response === "string" || !this.response.error)) {
       if (typeof this.response === "string" || Array.isArray(this.response)) {
+        //@ts-ignore
```
gcoonrod, line 111: Why is the ignore needed?

gcoonrod (commented):
Just some questions. No blocking change requests.

src/classes/actionProcessor.ts
evantahler: This one is because we typed `response` assuming it will be JSON (which it is in most cases) https://github.com/actionhero/actionhero/blob/main/src/classes/actionProcessor.ts#L26-L28

But, this if block is considering the case that response was set directly to be a string or something else.

src/classes/actionProcessor.ts
evantahler: I *think* that enums are more for objects - storing a reference key and a string value.  In this case, I just wanted to have a fixed collection of strings that could be used.

The enum way would be something like: 
```ts
enum ErrorStatusMessages {
  Complete: 'complete'
  GenericError: 'generic_error'
}

// this.completeAction("validator_errors") would become
this.completeAction(ErrorStatusMessages.Complete)
```

I don't really know which one is better?

evantahler: Ah ha! If you use enums, the values they resolve to are actually numeric if you don't supply a string value (like an object)

Screen Shot 2021-06-29 at 5 56 24 PM

Not that it matters too much, but perhaps using the strings is more expressive?

gcoonrod: I don't either. My java loving heart prefers Enums. But I have no technical reason to push for them.

gcoonrod (approved):

evantahler:
Let's use Enums! https://github.com/actionhero/actionhero/pull/1870/commits/51e84b3c8ffbfce770b99690bd9691a63946c5e3

gcoonrod (approved):

pedroslopez (approved):
Looks good!

PR #1931, opened 2021-08-24
Use `process.env.LOG_LEVEL` to set log level by default

evantahler:
New Actionhero projects will come with an environment variable, `process.env.LOG_LEVEL` which can be used to set the log level for both the file and console logger.

### DIFF
--- a/src/config/logger.ts
+++ b/src/config/logger.ts
@@ -3,6 +3,8 @@ import * as winston from "winston";
 /*
 The loggers defined here will eventually be available via `import { loggers } from "actionhero"`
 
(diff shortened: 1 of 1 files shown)


gcoonrod (approved):
Do we need to check that the env variable value is a valid choice?

evantahler:
> Do we need to check that the env variable value is a valid choice?

I think ideally we would, but you could have custom log levels too, and then we couldn't validate them :/

evantahler:
Added to the docs via https://github.com/actionhero/www.actionherojs.com/commit/37d0f81547c0e28933453055adf7549a2384932f

PR #1945, opened 2021-09-01
Added types for the loglevel

krishnaglick:
This may be a breaking change.
If the types include `string` then we no longer get the typescript autocomplete.

### DIFF
--- a/src/classes/action.ts
+++ b/src/classes/action.ts
@@ -1,5 +1,6 @@
-import { Inputs } from "./inputs";
 import { api } from "../index";
+import type { LogLevels } from "../modules/log";
+import { Inputs } from "./inputs";
 
 /**
  * Create a new Actionhero Action. The required properties of an action. These can be defined statically (this.name) or as methods which return a value.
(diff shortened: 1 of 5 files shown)


evantahler:
This is a good addition!

We use the [winston](https://github.com/winstonjs/winston) logger for Actionhero, and while the default loggers (created in [`src/config/loggers`](https://github.com/actionhero/actionhero/blob/main/src/config/logger.ts)) use these "syslog" style of log levels, you can change them to be other styles - like "npm" or even custom log levels - https://github.com/winstonjs/winston#logging-levels.

It also seems like the winston logger could provide these types for us:

```ts
import * as winston from 'winston'

export type LogLevel =
  | winston.config.SyslogConfigSetLevels
  | winston.config.CliConfigSetLevels;
```

To make this a non-breaking change, and to support folks changing their logger format, could this new `LogLevels` type be defined elsewhere... perhaps in the config file?

evantahler:
(doing a very unscientific survey in the Actionhero slack to see if folks use any log levels other than "syslog")

krishnaglick:
> I'm doing a very unscientific survey in the [Actionhero slack](https://slack.actionherojs.com) and Twitter to see if folks use any log levels other than "syslog". Let’s see if people change the log levels or not.

I feel like most just use info, error, or debug. Though I could be wrong!

gcoonrod (approved):

gcoonrod:
I think its a good idea to stick with the syslog levels.

evantahler:
OK, let's merge this! I'm getting crickets on twitter and slack.   

I'll make this a minor change (middle semver number), as I don't think this will be breaking for most people, but does indicate that something is different.

evantahler:
Documentation updates https://github.com/actionhero/www.actionherojs.com/pull/309

PR #1957, opened 2021-09-19
Rename `LogLevels` to `ActionheroLogLevel` and export

evantahler:
`ActionheroLogLevel` needs to be exported so consumers can type their logs


evantahler:
@krishnaglick FYI - the pattern we use is to have singular Types so that this variable is an instance of thing, e.g.: `let severity: ActionHeroLogLevel = 'info'` rather than  `let severity: ActionHeroLogLevels = 'info'`

PR #1977, opened 2021-10-01
`commander` v8

evantahler:


src/bin/methods/actions/list.ts
```diff
@@ -3,7 +3,7 @@ import { api, CLI } from "./../../../index";
 export class ActionsList extends CLI {
   constructor() {
     super();
-    this.name = "actions list";
+    this.name = "actions-list";
```
evantahler, line 6: This would be a breaking change, but this is not likely to be used by actionhero develoeprs

PR #2004, opened 2021-10-31
specHelper runAction runTask runFullTask generic type variable can be optional

WaBLin123:
I think it can be divided into two parts: generic type and return type from ```specHelper.runAction<A>()```

1. generic type from ```specHelper.runAction<A>()```
I choose to define the generic type as ```A extends Action | void = void```.  This allows to not provide type and if provide type, it must be ```Action```

- provide ```Action``` will be passed
image
- not provide type will be ```void``` and passed
image
- provide not ```Action```(provide ```Task```) will be rejected
image

2. return type from ```specHelper.runAction<A>()```
Providing ```Action``` will be the same as before. But if no type is provided(as provide ```void```), the "non-strict" type will be returned. The returned type should have some useful info(#2003 ), so I choose ```{[key: string]: any} & RequestExtraInfos```(For convenience, the type in the figure below will be called "RequestExtraInfos"), not ```any & RequestExtraInfos```,  because ```any & RequestExtraInfos``` will be ```any```.  This is not what we expected
image

And I have the same idea about ```specHelper.runTask``` and ```specHelper.runFullTask```, so I also make changes

### DIFF
--- a/src/modules/specHelper.ts
+++ b/src/modules/specHelper.ts
@@ -15,7 +15,7 @@ export namespace specHelper {
   /**
    * Run an action via the specHelper server.
    */
-  export async function runAction<A extends Action>(
+  export async function runAction<A extends Action | void = void>(
     actionName: string,
     input: { [key: string]: any } = {}
   ) {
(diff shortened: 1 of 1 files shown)


evantahler:
This is great @WaBLin123, thank you!

There's a question to answer about the philosophy of Actionhero's testing library - how strongly do we want to encourage that folks use Types?  Do we want to strongly encourage testing with types (the behavior before this PR) or do we want to allow folks to "opt-into" Typed behavior.   I think I agree with this PR, and that we should be allowing folks to Opt-into Typed-testing, rather than require it.... for now anyway. 

@gcoonrod - what do you think?

evantahler (approved):

gcoonrod:
I think this is a good idea.

evantahler:
Thanks again @WaBLin123!

evantahler:
Released! https://github.com/actionhero/actionhero/releases/tag/v27.2.1

PR #2012, opened 2021-11-12
Simplify the arguments to action formatters and validators

evantahler:
This PR simplifies the arguments Actions pass to input Default, Formatters, and Validators.  Before this PR a second argument of `this` would be passed, which was the `api` object for some reason.  Now that accessing the `api` is object is as simple as `import {api} from 'actionhero'` this isn't needed.  This means that there are no un-expected inputs to your input methods.  

This PR also clarifies that the `this` scope of input methods is the Action Processor itself - so you'll have access to the action's class, the connection, etc - the same as `data` on the Action' s run method if you need it though `this.connection` and `this.actionTemplate`. 

This is technically a breaking change, although I don't think that was used...

### DIFF
--- a/src/classes/actionProcessor.ts
+++ b/src/classes/actionProcessor.ts
@@ -257,7 +257,7 @@ export class ActionProcessor<ActionClass extends Action> {
     // default
     if (params[key] === undefined && props.default !== undefined) {
       if (typeof props.default === "function") {
-        params[key] = await props.default.call(api, params[key], this);
+        params[key] = await props.default.call(this, params[key]);
       } else {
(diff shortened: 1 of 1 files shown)


evantahler:
Docs for the change in https://github.com/actionhero/www.actionherojs.com/pull/331

evantahler:
cc @krishnaglick

krishnaglick (approved):
Probably won't break everything! 😃

evantahler:
I'll make this a minor-version release (vs major or patch) to indicate that *something* has changed

PR #2014, opened 2021-11-12
No Implicit Any

evantahler:
🎉 🎉 🎉  All-in on Typescript!  🎉 🎉 🎉 

This Pull Request will remove any implicit "any" types from our project, and requires that all future contributions are well-typed. This has the huge advantage of creating and exporting Typescript types for consuming projects to use. This is a big undertaking and will change a number of APIs (for the better).

Features:
* The `config` object is now typed!
* Module declarations are now auto-typed from their implementation.

Screen Shot 2021-11-12 at 5 30 23 PM
Screen Shot 2021-11-12 at 5 31 01 PM

Breaking Changes
* `config.servers` is gone.  `config.servers.web` -> `config.web` and `config.servers.websocket` -> `config.websocket`
* Config files have a new format and declare their config namespace against the actionhero module
* You cannot change the `config` object any more with an override when starting the process.  You can use `ACTIONHERO_CONFIG_OVERRIDES` to set some overrides in the ENV

PR #2022, opened 2021-11-23
Fix cluster import

evantahler:
The changes made in https://github.com/actionhero/actionhero/pull/2014 around the cluster module need to be reverted.  The `@types` for this package are incorrect.  Crazy!

```ts
// not working

import cluster from "cluster";
console.log(cluster); // --> undefined

// working
import * as cluster from "cluster";
console.log(cluster); // --> 

EventEmitter {
  _events: [Object: null prototype] {},
  _eventsCount: 0,
  _maxListeners: undefined,
  isWorker: false,
  isMaster: true,
  isPrimary: true,
  Worker: [Function: Worker],
  workers: {},
  settings: {},
  SCHED_NONE: 1,
  SCHED_RR: 2,
  schedulingPolicy: 2,
  setupPrimary: [Function (anonymous)],
  setupMaster: [Function (anonymous)],
  fork: [Function (anonymous)],
  disconnect: [Function (anonymous)],
  [Symbol(kCapture)]: false
}
```

### DIFF
--- a/src/classes/process/id.ts
+++ b/src/classes/process/id.ts
@@ -1,4 +1,4 @@
-import cluster from "cluster";
+import * as cluster from "cluster";
 import { config } from "./../../modules/config";
 import { utils } from "../../modules/utils";
 
@@ -21,6 +21,7 @@ function determineId() {
     }
 
     id = externalIP;
+    // @ts-ignore - we need to load * as for node v16 support
     if (cluster["isWorker"]) id += `:${process.pid}`;
   } else {
     id = config.general.id;


evantahler:
cc @krishnaglick - nuts!

krishnaglick:
Probably worth adding `esModuleInterop` to tsconfig

evantahler:
I don’t think that was the issue though - in this case the types provided by @types/node were wrong. Usually, with the other builtins like fs, path, etc know about the different import styles

PR #2024, opened 2021-11-25
Fixes router bug matching variable routes without a required prefix

evantahler:
This PR fixes a bug with the Actionhero Router.

Prior to this PR, a route like `/v:apiVersion/status` would match `/1/status` when it shouldn't.  The expected matches should have a `"v"` in the routes, like `/v1/status`

### DIFF
--- a/src/initializers/routes.ts
+++ b/src/initializers/routes.ts
@@ -91,7 +91,7 @@ export class RoutesInitializer extends Initializer {
       params: {},
     };
     const matchParts = match.split("/");
-    let regexp = "";
+    let regexpMatch: string = null;
     let variable = "";
 
     if (matchParts[0] === "") matchParts.splice(0, 1);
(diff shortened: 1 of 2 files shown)


src/initializers/routes.ts
```diff
@@ -105,33 +105,34 @@ export class RoutesInitializer extends Initializer {
...
       if (!pathPart) return response;
 
-      if (matchPart.indexOf(":") >= 0) {
+      if (matchPart.includes(":")) {
         const trimmedMatchParts = matchPart.split(":");
         const trimmedMatchPart =
           trimmedMatchParts[trimmedMatchParts.length - 1];
         const replacement = trimmedMatchParts[trimmedMatchParts.length - 2];
         if (replacement) {
+          if (!pathPart.includes(replacement)) return response;
```
evantahler, line 122: This is the fix - we weren't exiting if there was nothing to match, so a "pathPart" or `v1` would be replaced to be `1`... the same as if it was just `1` to begin with.  We need to not match if there was no "v" in the first place

tealjulia (approved):

krishnaglick (approved):

gcoonrod (approved):

PR #2039, opened 2021-12-03
use global flag in swagger path substitution

pedroslopez:
Allows replacement of more than 5 variables in the path by using the global flag in the regex. This works because technically `$1` is still the first capture group in the provided regex.


evantahler:
^ This @gitpod thing is cool! First time trying it out

evantahler (approved):

evantahler:
Released! https://github.com/actionhero/actionhero/releases/tag/v28.0.2

PR #2052, opened 2021-12-17
Do not assume that all uses of Actionhero in test are running typescript

evantahler:
We need to know if Actionhero is running in Typescript (via `ts-node`, `ts-node-dev`, `jest`, or similar) to determine if we should load `/src/config/*.ts` or `/dist/config/*.js`

Prior to this PR, we assumed that any time that `NODE_ENV=test`, we should be running typescript and load the typescript config files.  This prevented anyone from using a CLI Actionhero command in CI.  This new way of guessing if we are running in typescript should be a bit safer.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "actionhero",
-  "version": "28.0.3",
+  "version": "28.0.4-alpha.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "actionhero",
-      "version": "28.0.3",
+      "version": "28.0.4-alpha.0",
       "hasInstallScript": true,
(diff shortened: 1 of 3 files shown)


evantahler:
@gcoonrod do you have any memory of why we may have wanted for force TS usage in test?  I can't :(

gcoonrod:
@evantahler I'd guess we just didn't think about why we _wouldn't_ want to run TS everywhere.

gcoonrod (approved):

evantahler:
> @evantahler I'd guess we just didn't think about why we _wouldn't_ want to run TS everywhere.

Ha! Yeah.  ...I'm not really sure how to test this PR since I don't really know what the feature was to begin with :(

PR #2053, opened 2021-12-17
Enhance Typing config in logger

krishnaglick:
Better types for the loggers we build in `/src/config/logger.ts`


evantahler:
Thanks @krishnaglick!  Looks like this version is better typed than our internal usage :D

PR #2059, opened 2021-12-21
`ParamsFrom` utility to determine types of params from Action Inputs

evantahler:
The new `ParamsFrom` TS utility can *finally* help dynamically determine the param types of an Action's run method!  This only works on Action classes that **statically** define the inputs, e.g. NOT in a constructor.  Here's a screenshot of things working:

Screen Shot 2021-12-20 at 7 14 01 PM

Note that `params.value` is properly a string, and we get an error that `params.foo` is undefined.

Here's the magic:
```ts
export type ParamsFrom<A extends Action | Task> = {
  [Input in keyof A["inputs"]]: A["inputs"][Input]["formatter"] extends (
    ...ags: any[]
  ) => any
    ? ReturnType<A["inputs"][Input]["formatter"]>
    : string;
};
```

What's nice about this approach is that we don't need to change any of Actionhero's internals, but we should start writing all of our Actions and Tasks with static definitions.

### DIFF
--- a/src/actions/cacheTest.ts
+++ b/src/actions/cacheTest.ts
@@ -1,39 +1,36 @@
-import { cache, Action } from "./../index";
+import { cache, Action, ParamsFrom } from "./../index";
 
 export class CacheTest extends Action {
-  constructor() {
-    super();
-    this.name = "cacheTest";
-    this.description = "I will test the internal cache functions of the API";
-    this.inputs = {
-      key: {
-        required: true,
-        formatter: this.stringFormatter,
-        validator: this.stringValidator,
-      },
+  name = "cacheTest";
+  description = "I will test the internal cache functions of the API";
+  inputs = {
+    key: {
+      required: true,
+      formatter: this.stringFormatter,
+      validator: this.stringValidator,
+    },
 
-      value: {
-        required: true,
-        formatter: this.stringFormatter,
-        validator: this.stringValidator,
-      },
-    };
+    value: {
+      required: true,
+      formatter: this.stringFormatter,
+      validator: this.stringValidator,
+    },
+  };
 
-    this.outputExample = {
-      cacheTestResults: {
-        saveResp: true,
-        sizeResp: 1,
-        loadResp: {
-          key: "cacheTest_key",
-          value: "value",
-          expireTimestamp: 1420953274716,
-          createdAt: 1420953269716,
-          readAt: null,
-        },
-        deleteResp: true,
+  outputExample = {
+    cacheTestResults: {
+      saveResp: true,
+      sizeResp: 1,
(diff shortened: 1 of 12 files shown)


evantahler:
Sadly, because the run method /implements/ the method from the abstract class, we cannot pre-determine the type or params within the ActionProcessor :(

src/classes/inputs.ts
```diff
@@ -1,5 +1,15 @@
 import { Input } from "./input";
+import { Action } from "../classes/action";
+import { Task } from "../classes/task";
 
 export interface Inputs {
   [key: string]: Input;
 }
+
+export type ParamsFrom<A extends Action | Task> = {
```
gcoonrod, line 9: I don't love this syntax. Gives me flashbacks to Reflections in Java. I don't know TS well enough to critique this one way or the other though.

src/tasks/runAction.ts
```diff
@@ -1,22 +1,18 @@
...
+  queue = "default";
 
-  async run(params: Record<string, any>) {
+  async run(params: ParamsFrom<RunAction>) {
     if (!params) params = {};
 
     const response = await action.run(
       params.action,
       params.version,
+      // @ts-ignore
```
gcoonrod, line 15: Why is this needed?

gcoonrod (approved):
Overall I think this is a good idea. My TS knowledge isn't enough to speak authoritatively on the implementation though.

src/actions/cacheTest.ts
```diff
@@ -1,39 +1,36 @@
...
+        value: "value",
+        expireTimestamp: 1420953274716,
+        createdAt: 1420953269716,
+        readAt: null as number,
       },
-    };
-  }
+      deleteResp: true,
+    },
+  };
```
krishnaglick, line 33: So we need to define these properties at the top-level in the class, rather than in the constructor to make it work? That seems sane.

src/classes/inputs.ts
krishnaglick: Unfortunately it's the reality of Typescript that shit's going to get weird like this when you infer complex types. Theoretically `params: typeof this.params` should work but it doesn't.

src/tasks/runAction.ts
krishnaglick: Do Tasks need to be:
```
async run(params: {
    action: string;
    version: string;
    params: ParamsFrom<RunAction>;
  }) {
    const response = await action.run(
      params.action,
      params.version,
      params.params
    );
```
Maybe?

src/actions/cacheTest.ts
evantahler: Yep.

PR #2076, opened 2022-01-06
Rebuild Config on Server Restart

krishnaglick:
Adding rebuild config method. Calling rebuild config method on server restart.

### DIFF
--- a/__tests__/modules/config.ts
+++ b/__tests__/modules/config.ts
@@ -1,4 +1,5 @@
 import { config } from "../../src/index";
+import { rebuildConfig } from "../../src/modules/config";
 
 describe("config class", () => {
   it("loads config", () => {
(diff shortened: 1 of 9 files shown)


evantahler (changes requested):
This is probably important to test.

Here's a potential path forward:
1. set a config with an ENV (like PORT=12345)
1. startup the test like normal
https://github.com/actionhero/actionhero/blob/80895afc49f62b6842033b98f6e89ae7ad02ca26/__tests__/actions/cacheTest.ts#L5-L7
2. test that the config is what you expect (config.web.port   === 12345)
3. change the ENV in the test
4. assert that the config is the old value
5. reboot the server (process.restart)
6. assert that the config is the new value

__tests__/modules/config.ts
```diff
@@ -7,4 +8,12 @@ describe("config class", () => {
...
     expect(config.web.httpHeaders["X-Powered-By"]).toBe("actionhero");
   });
+
+  it("updates config on 'restart'", () => {
+    const port = "2020";
+    process.env.PORT = port;
+    expect(config.web.port).not.toBe(port);
+    rebuildConfig();
+    expect(config.web.port).toBe(port);
+  });
```
evantahler, line 18: 👍

src/modules/config.ts
```diff
@@ -222,4 +231,12 @@ export function buildConfig() {
...
-export const config = buildConfig();
+export let config = buildConfig();
+export const rebuildConfig = () => {
+  recalculateEnv();
+  recalculateActionheroVersion();
+  recalcuateId();
+  recalculateProjectRoot();
+  recalculateIsTypescript();
+  config = buildConfig();
+};
```
evantahler, line 242: This is a great pattern!

evantahler (approved):

PR #2099, opened 2022-01-27
use commander to handle cli `formatter`/`validator`

pedroslopez:
Moves to using commander's [built-in option processing](https://github.com/tj/commander.js#custom-option-processing) when running `formatter` and `validator` functions on CLI inputs.

This lets commander give us better output in the case of validation errors (as opposed to showing a full stack trace as currently happens).

```
▲ actionhero % actionhero something --test blah.
error: option '--test [test]' argument 'blah.' is invalid. blah. isnt valid because periods are bad
```

**Possibly breaking**: `formatter` and `validator` can no longer be asynchronous for CLI inputs.

### DIFF
--- a/__tests__/cli/cli.ts
+++ b/__tests__/cli/cli.ts
@@ -1,14 +1,6 @@
 import * as ChildProcess from "child_process";
 import * as path from "path";
 
-process.env.ACTIONHERO_CONFIG_OVERRIDES = JSON.stringify({
-  general: {
-    paths: {
-      cli: [path.join(__dirname, "..", "testCliCommands")],
-    },
-  },
-});
-
 async function exec(
   command: string,
   args: Record<string, any>
@@ -26,7 +18,17 @@ async function exec(
 }
 
 describe("cli commands", () => {
-  const env = process.env;
+  const env = {
+    ...process.env,
+    ACTIONHERO_CONFIG_OVERRIDES: JSON.stringify({
+      general: {
+        paths: {
+          cli: [path.join(__dirname, "..", "testCliCommands")],
+        },
+      },
+    }),
+  };
+
   test(
     "new commands appear in help",
     async () => {
--- a/__tests__/cli/cli.ts
+++ b/__tests__/cli/cli.ts
@@ -1,0 +1,109 @@
+import * as ChildProcess from "child_process";
+import * as path from "path";
+
+process.env.ACTIONHERO_CONFIG_OVERRIDES = JSON.stringify({
+  general: {
+    paths: {
+      cli: [path.join(__dirname, "..", "testCliCommands")],
+    },
+  },
+});
+
+async function exec(
+  command: string,
+  args: Record<string, any>
+): Promise<{
+  error?: NodeJS.ErrnoException;
+  stdout?: string;
+}> {
+  return new Promise((resolve, reject) => {
+    ChildProcess.exec(command, args, (error, stdout, stderr) => {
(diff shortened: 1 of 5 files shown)


src/bin/actionhero.ts
```diff
@@ -124,7 +124,37 @@ export namespace ActionheroCLIRunner {
...
+
+          if (typeof input.validator === "function") {
+            input.validator(value);
+          }
+
+          if (input.variadic) {
+            if (!Array.isArray(accumulator)) accumulator = [];
+            accumulator.push(value);
+            return accumulator;
+          }
```
pedroslopez, line 144: Turns out that adding this custom processor disables the default behavior of accumulating variadic options and returning them as an array, so had to add it back in.

See https://github.com/tj/commander.js/blob/43f4743864e2f670db5eebcf88c92aa4612c54f1/lib/command.js#L543-L555

__tests__/testCliCommands/hello.ts
```diff
@@ -0,0 +1,41 @@
...
+      formatter: (val: string) => {
+        return `${val}.`;
+      },
+      validator: (val: string) => {
+        const parts = val.split(".");
+        if (parts.length > 2) throw new Error("too many periods");
+      },
+    },
+    countries: {
+      variadic: true as const,
```
pedroslopez, line 24: For CLI inputs, if this is a variadic option we always return an array (of `formatter`'s return type or `string`).
To make this work properly unfortunately `variadic` needs to be set as a constant when being defined for an input.

pedroslopez:
@evantahler I thought I may have broken something here but it looks like tests on main are failing as well after the `ioredis-mock` upgrade? 🤔

src/bin/actionhero.ts
evantahler: Facinating!

evantahler:
> @evantahler I thought I may have broken something here but it looks like tests on main are failing as well after the `ioredis-mock` upgrade? #2094 🤔

Looking into that now...

__tests__/cli/cli.ts
```diff
@@ -0,0 +1,111 @@
...
+      const { stdout } = await exec(
+        "./node_modules/.bin/ts-node ./src/bin/actionhero.ts hello --title Mr --name Worldwide --countries France Italy Germany USA",
+        { env }
+      );
+      expect(stdout).toContain(
+        "Hello, Mr. Worldwide (France! Italy! Germany! USA!)"
+      );
+    },
+    30 * 1000
+  );
```
evantahler, line 95: 🕴️ 🗺️

src/classes/inputs.ts
```diff
@@ -7,10 +7,20 @@ export interface Inputs {
...
 }
 
-export type ParamsFrom<A extends Action | Task | CLI> = {
-  [Input in keyof A["inputs"]]: A["inputs"][Input]["formatter"] extends (
-    ...ags: any[]
-  ) => any
-    ? ReturnType<A["inputs"][Input]["formatter"]>
-    : string;
-};
+export type ParamsFrom<A extends Action | Task | CLI> = A extends CLI
```
evantahler, line 10: I had a feeling that this type was going to get nuts - that's why I wanted to build it into the project so we can maintain it for others

evantahler (approved):
I'll get the main branch green again (ioredis) and then this can be merged in!

pedroslopez:
@evantahler looks like we're green now! I'll let you merge

PR #2106, opened 2022-01-28
Bump node-resque and ioredis-mock

evantahler:


evantahler:
There's a bug in newer versions of ioredis-mock that does strange things after calling `flushdb` (https://github.com/stipsan/ioredis-mock/issues/1124)... so, for now, we just won't use that in our tests :/

src/config/logger.ts
```diff
@@ -65,7 +65,7 @@ function buildConsoleLogger(level = "info") {
           } ${stringifyExtraMessagePropertiesForConsole(info)}`;
         })
       ),
-      level,
+      level: level as winston.level,
```
evantahler, line 68: Looks like the winston types changed a bit too...

PR #2108, opened 2022-01-28
fix `ParamsFrom` type inference

pedroslopez:
For some reason, TypeScript has trouble inferring types properly when using `A extends CLI` as a conditional.

image

You can play around with this example here https://github.com/actionhero/actionhero/commit/eee6e8dcab5bcfe795fe70c262f1d9f7b6909ab3

Separating out ParamsFrom for the CLI to it's own type lets you explicitly use it, though I'd still like to find a solution that keeps things in a single `ParamsFrom` that works for both if possible, hence the draft PR. Pushed here for visibility.

### DIFF
--- a/__tests__/testCliCommands/hello.ts
+++ b/__tests__/testCliCommands/hello.ts
@@ -1,4 +1,4 @@
-import { CLI, ParamsFrom } from "./../../src/index";
+import { CLI, CLIParamsFrom } from "./../../src/index";
 
 export class HelloCliTest extends CLI {
   name = "hello";
@@ -31,7 +31,7 @@ export class HelloCliTest extends CLI {
     },
   };
 
-  async run({ params }: { params: Partial<ParamsFrom<HelloCliTest>> }) {
+  async run({ params }: { params: Partial<CLIParamsFrom<HelloCliTest>> }) {
     console.log(
       `Hello, ${params.title} ${params.name} ${
         params.countries ? `(${params.countries.join(" ")})` : ""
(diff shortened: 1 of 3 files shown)


__tests__/testCliCommands/hello.ts
```diff
@@ -22,7 +22,7 @@ export class HelloCliTest extends CLI {
       },
     },
     countries: {
-      variadic: true as const,
+      variadic: true as true,
```
krishnaglick, line 25: I was able to consolidate the Params type, but `as const` or `as true` is still required since typescript can't narrow the type well enough itself.

evantahler:
Oh, this is a nice improvement to `ParmsFrom` either way!

evantahler:
Oh, this is a nice improvement to `ParmsFrom` either way!

evantahler (approved):

__tests__/testCliCommands/hello.ts
pedroslopez: Nice!

krishnaglick: What we could also do is make a function called VariadicType that just takes an Input and says what our type is doing. Then we could avoid the `as const` part.

src/classes/inputs.ts
```diff
@@ -7,20 +7,15 @@ export interface Inputs {
...
-        : string;
-    };
+type FormatterOrString<I extends (Action | Task | CLI)["inputs"][string]> =
+  I["formatter"] extends (...args: any[]) => any
+    ? ReturnType<I["formatter"]>
+    : string;
+
+type Variadic = { variadic: true };
+
+export type ParamsFrom<A extends Action | Task | CLI> = {
```
krishnaglick, line 17: It might be worth putting `Action | Task | CLI` into a single type, but I don't think this is likely to change.

evantahler (approved):

evantahler:
Releasing a new version ow

PR #2130, opened 2022-02-23
Added new ParamsFrom so requests utilizing ParamsFrom don't require path params in the request body.

krishnaglick:
Unsure about naming, but going with Request since this is to allow input inference without requiring url params in the payload

### DIFF
--- a/src/classes/inputs.ts
+++ b/src/classes/inputs.ts
@@ -18,8 +18,17 @@ type RequiredParamsKeys<A extends ActionheroWithParams> = KeysOfType<
   A["inputs"],
   Required
 >;
+
 type Variadic = { variadic: true };
 type Required = Readonly<{ required: true }> | { required: true };
+
+type RequestRequiredParamsKeys<A extends ActionheroWithParams> = KeysOfType<
+  A["inputs"],
+  RequestRequired
+>;
+type RequestRequired =
+  | Readonly<{ required: true; inPath?: false }>
+  | { required: true; inPath?: false };
 
 type ParamsExtractor<A extends ActionheroWithParams> = {
   [Input in keyof A["inputs"]]: A["inputs"][Input] extends Variadic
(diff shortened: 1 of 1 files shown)


evantahler (commented):
I'm struggling with this one.  As far as the Action is concerned, the param really is required and should be there in the `run()` method.  The `ParamsFrom<>` type is primarily for consuming within the Action itself.  This addition is for clients who want to know a TS type about what input the Action needs, outside of Actionhero.  I think that *might* be the concern of the client consuming actionhero.  For example, maybe in Grouparoo, we know that all routes only would ever have the param `id`, so maybe we can mix the type of `ParamsFrom<ActionType>` with `Partial<{id?: string>`?

Should we want to move forward with this:

Some alternatives names:
* `sourcedIndirectly` (could be from a URL, hostname, etc)
* `primaryKey` (maybe being a primary key implies the param is always in the URL?)

To merge this, we would need an example that uses this new option and some updates to the www.actionherojs.com docs.

krishnaglick:
> I'm struggling with this one. As far as the Action is concerned, the param really is required and should be there in the `run()` method. The `ParamsFrom<>` type is primarily for consuming within the Action itself. This addition is for clients who want to know a TS type about what input the Action needs, outside of Actionhero. I think that _might_ be the concern of the client consuming actionhero. For example, maybe in Grouparoo, we know that all routes only would ever have the param `id`, so maybe we can mix the type of `ParamsFrom<ActionType>` with `Partial<{id?: string>`?
> 
> Should we want to move forward with this:
> 
> Some alternatives names:
> 
>     * `sourcedIndirectly` (could be from a URL, hostname, etc)
> 
>     * `primaryKey` (maybe being a primary key implies the param is always in the URL?)
> 
> 
> To merge this, we would need an example that uses this new option and some updates to the [www.actionherojs.com](http://www.actionherojs.com) docs.

I don't mind making this the concern of the client. Happy to do that if you'd prefer!
Otherwise I think `sourcedIndirectly` might be the way to go.

PR #2196, opened 2022-05-27
fix websocket bind logger message

rakhnin:


evantahler:
Thanks!

Before this PR the log line would have been:

```
2022-05-27T15:56:20.361Z - debug: [server: websocket] webSockets bound to undefined: undefined
```

PR #2276, opened 2022-08-30
Update redis and redis packages

evantahler:
For now, we need to remove `@types/ioredis` because they rely on the internal types of a previous version of `@types/ioredis`


evantahler:
This unlocks merging https://github.com/actionhero/actionhero/pull/2241

PR #2541, opened 2023-07-14
remove readAt and expireTimestamp from cache objects

### LINKED ISSUE #2540: [Bug]: Cache inconsistency
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

If two Actions are working with the same value in the cache at the same time, it's possible to get some inconsistent behavior. I was load testing a new application I wrote with Actionhero and was able to reproduce this fairly regularly. 

Example:
* The cache holds the value `0` at `key`
* Action A saves `1` to `key`
* At almost exactly the same time, Action B loads `key` from the cache
* B gets `0` back just before A saves `1`
* A second later, loading `key` from the cache returns `0`

### Expected Behavior

I believe this line should be removed: https://github.com/actionhero/actionhero/blob/main/src/modules/cache.ts#L219. With that in place, it's possible for a call to `cache.save` to resolve and then have the value it just saved overwritten immediately by another call to `cache.load` on the same node. 

I'd be happy to open a PR for this, if that helps. I tried to join the Actionhero Slack, but https://slack.actionherojs.com/ is not responding.

### Stack Trace

_No response_

### Environment

OSX
Actionhero version 28.3.3
Node.js version 18.16.0

### Additional context

_No response_

rogorman9:
Fixes #2540

### DIFF
--- a/src/modules/cache.ts
+++ b/src/modules/cache.ts
@@ -5,16 +5,12 @@ export namespace cache {
   export enum CacheErrorMessages {
     locked = "Object locked",
     notFound = "Object not found",
-    expired = "Object expired",
   }
 
   export interface CacheObject {
     key: string;
     value: any;
-    expireTimestamp: number;
     createdAt: number;
-    lastReadAt: number;
-    readAt?: number;
   }
 
   export interface CacheOptions {
@@ -152,14 +148,7 @@ export namespace cache {
     const count = Object.keys(data).length;
 
     const saveDumpedElement = async (key: string, content: any) => {
-      const parsedContent = JSON.parse(content);
-      await client().set(key, content);
-      if (parsedContent.expireTimestamp) {
-        const expireTimeSeconds = Math.ceil(
-          (parsedContent.expireTimestamp - new Date().getTime()) / 1000,
-        );
-        await client().expire(key, expireTimeSeconds);
-      }
+      await client().set(key, content, "KEEPTTL");
     };
 
     Object.keys(data).forEach((key) => {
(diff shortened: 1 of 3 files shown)


src/modules/cache.ts
```diff
@@ -191,48 +180,21 @@ export namespace cache {
...
-    }
-
     lockOk = await cache.checkLock(key, options.retry);
     if (lockOk !== true) throw new Error(CacheErrorMessages.locked);
 
-    await client().set(redisPrefix + key, JSON.stringify(cacheObj));
-    if (expireTimeSeconds) {
-      await client().expire(redisPrefix + key, expireTimeSeconds);
+    if (options.expireTimeMS) {
+      await client().pexpire(redisPrefix + key, options.expireTimeMS);
```
evantahler, line 187: oh yay, we got an expire in ms method!

evantahler (commented):
This looks good!
As none of the method signatures changed, but the body content of the cached objects did... do you think this should be a breaking change or not 🤔 ?

evantahler:
cc @gcoonrod - what are you thoughts on if this would be a breaking change or not?

gcoonrod:
> cc @gcoonrod - what are you thoughts on if this would be a breaking change or not?

Yes, it's a change to the members of the pubic `CacheObject` API, which I can imagine some projects might have used.

gcoonrod (approved):

evantahler:
Related PRs for the version bump https://github.com/actionhero/actionhero/pull/2542 and docs update https://github.com/actionhero/www.actionherojs.com/pull/456

PR #2595, opened 2023-09-13
add removeMember to api.chatRoom

rogorman9:
`removeMember` can be called as an RPC method [here](https://github.com/actionhero/actionhero/blob/main/src/modules/chatRoom.ts#L273). Since `removeMember` is not on the `api.chatRoom` object, it does not get executed and just logs `RPC method chatRoom.removeMember not found`.

### DIFF
--- a/__tests__/core/cluster.ts
+++ b/__tests__/core/cluster.ts
@@ -101,6 +101,20 @@ describe("Core: Action Cluster", () => {
       expect(message.message).toEqual("hi");
     });
 
+    test("can call api.chatRoom.removeMember on other servers", async () => {
+      const client = await specHelper.buildConnection();
+      const spy = jest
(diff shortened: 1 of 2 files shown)


evantahler:
Great find! 
Is there a test you can add that was missing for this?

evantahler:
Also, please rebase on the latest `main` branch - https://github.com/actionhero/actionhero/pull/2596 fixed those test failures

rogorman9:
@evantahler Just rebased and added a test. Hopefully that covers it, but let me know if there's a better place/method for testing these.

evantahler:
That's a fine test! Thank you!

evantahler:
Published in https://github.com/actionhero/actionhero/releases/tag/v29.0.3

PR #2640, opened 2023-11-02
Pass options to form contructor

321hendrik:
As described [here](https://github.com/node-formidable/formidable#formidable--incomingform) options for IncomingForm should be passed to the contructor directly.

### DIFF
--- a/src/servers/web.ts
+++ b/src/servers/web.ts
@@ -693,7 +693,13 @@ export class WebServer extends Server {
         (connection.rawConnection.req.headers["content-type"] ||
           connection.rawConnection.req.headers["Content-Type"])
       ) {
-        connection.rawConnection.form = new formidable.IncomingForm(this.config.formOptions);
+        connection.rawConnection.form = new formidable.IncomingForm();
+        if (this.config?.formOptions) {
+          for (i in this.config.formOptions) {
+            connection.rawConnection.form.options[i] =
+              this.config.formOptions[i];
+          }
+        }
 
         let rawBody = Promise.resolve(Buffer.alloc(0));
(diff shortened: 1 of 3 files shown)


src/servers/web.ts
```diff
@@ -693,10 +693,7 @@ export class WebServer extends Server {
         (connection.rawConnection.req.headers["content-type"] ||
           connection.rawConnection.req.headers["Content-Type"])
       ) {
-        connection.rawConnection.form = new formidable.IncomingForm();
-        for (i in this.config.formOptions) {
-          connection.rawConnection.form.options[i] = this.config.formOptions[i];
-        }
+        connection.rawConnection.form = new formidable.IncomingForm(this.config.formOptions);
```
evantahler, line 696: Good catch! I like this change on principal.  

since this value is coming from [config](https://github.com/actionhero/actionhero/blob/main/src/config/web.ts#L81-L86), there is a chance it could be null or undefined, or even an empty object.  Could you test for that?

evantahler (changes requested):

evantahler:
Also, note the linter error

evantahler:
Ping! Are you still interested in working on this?

321hendrik:
Sorry, I couldn't spend time on this for a while. I made the requested changes and also added a type to the options for code intellisense.

src/servers/web.ts
321hendrik: I added a check and fixed the linter error.

evantahler:
Released in https://github.com/actionhero/actionhero/releases/tag/v29.1.0

PR #2641, opened 2023-11-02
Allow custom content-type header for sending files

321hendrik:
Makes it possible to configure the content-type header before sending a file. 

Use case:
- There is a file named `test.png_preview`
- Mimetype can not be inferred from file extension
- You want to specify the type yourself in the download action

### DIFF
--- a/__tests__/integration/sendFile.ts
+++ b/__tests__/integration/sendFile.ts
@@ -21,4 +21,9 @@ describe("Server: sendFile", () => {
     expect(stats.size).toBeGreaterThanOrEqual(response.data.length);
     expect(response.data).toContain("PNG");
   });
+
+  test("Server should sendFile with custom content-type header", async () => {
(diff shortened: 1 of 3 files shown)


evantahler:
I understand your use-case, and I agree we should allow for a way to make this change.  I think that I would prefer the server/action be in control of if we want to set a custom mimeType rather than relying on a user request (which could be anything), e.g.:

```js
await data.connection.sendFile("logo/actionhero.png", {mimeType: "application-foo"});
```

If the `sendFile` method were changed instead, then application could verify that the mimetype that is was about to set was valid/reasonable vs just passing back whatever the user requested... which could get folks into trouble (e.g. `application/x-msdownload`)

evantahler:
Ping! Are you still interested in working on this?

321hendrik:
I tried to implement your suggestion. Is this how you imagined?

evantahler:
Hello! Checking in again to see if you need help

PR #2699, opened 2024-01-15
Allow validators to have access to their param name

thmisch:
## Issue
Currently, there is no way to use generic validators that include the name of the invalid parameter in the error message.

Expected functionality:
```ts
function validateXYZ(param: string, name: string) {
	if (...) {
		throw new Error(`Parameter ${name} is of type ${...} (expected ${...})`)
	}
} 
```

This PR allows for this by passing `key` as extra parameter. Also, it's backwards compatible as additional arguments are ignored.

### DIFF
--- a/__tests__/actions/validationTest.ts
+++ b/__tests__/actions/validationTest.ts
@@ -28,6 +28,28 @@ describe("Action: validationTest", () => {
     );
   });
 
+  test("fails with generalized validation error message", async () => {
+    const { error } = await specHelper.runAction<ValidationTest>(
+      "validationTest",
+      {
+        string: "hello",
+        number: "invalid number",
+      },
+    );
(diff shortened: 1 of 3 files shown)


evantahler:
This is a great change! 

For this to be mergable, please:
1. Add tests!  Perhaps you can modify the [validationTest](https://github.com/actionhero/actionhero/blob/main/src/actions/validationTest.ts) actiom and related [tests](https://github.com/actionhero/actionhero/blob/main/__tests__/actions/validationTest.ts)
2. Open a related PR on the documentation website to share the changes

Once this is merged in, I'll release a new minor change for Actionhero

evantahler:
I also think the formatter should gain the same enhancement.

thmisch:
docs update: https://github.com/actionhero/www.actionherojs.com/pull/475

